Issue #14, opened 2016-12-09
User email address validation problem


mbag:
Characters allowed in email address field when registering are not the same as characters checked by url regex. 
Definitive solution TBD in comments.

```
Traceback (most recent call last):

  File "/opt/erigones/envs/lib/python2.7/site-packages/django/core/handlers/base.py", line 112, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)

  File "/opt/erigones/envs/lib/python2.7/site-packages/django/contrib/auth/decorators.py", line 22, in _wrapped_view
    return view_func(request, *args, **kwargs)

  File "/opt/erigones/gui/decorators.py", line 24, in wrap
    return fun(request, *args, **kwargs)

  File "/opt/erigones/gui/decorators.py", line 53, in wrap
    return fun(request, *args, **kwargs)

  File "/opt/erigones/gui/dc/user/views.py", line 67, in dc_user_list
    return render(request, 'gui/dc/user_list.html', context)

  File "/opt/erigones/envs/lib/python2.7/site-packages/django/shortcuts/__init__.py", line 53, in render
    return HttpResponse(loader.render_to_string(*args, **kwargs),

  File "/opt/erigones/envs/lib/python2.7/site-packages/django/template/loader.py", line 169, in render_to_string
    return t.render(context_instance)

  File "/opt/erigones/envs/lib/python2.7/site-packages/django/template/base.py", line 140, in render
    return self._render(context)

  File "/opt/erigones/envs/lib/python2.7/site-packages/django/template/base.py", line 134, in _render
    return self.nodelist.render(context)

  File "/opt/erigones/envs/lib/python2.7/site-packages/django/template/base.py", line 840, in render
    bit = self.render_node(node, context)

  File "/opt/erigones/envs/lib/python2.7/site-packages/django/template/base.py", line 854, in render_node
    return node.render(context)

  File "/opt/erigones/envs/lib/python2.7/site-packages/django/template/loader_tags.py", line 123, in render
    return compiled_parent._render(context)

  File "/opt/erigones/envs/lib/python2.7/site-packages/django/template/base.py", line 134, in _render
    return self.nodelist.render(context)

  File "/opt/erigones/envs/lib/python2.7/site-packages/django/template/base.py", line 840, in render
    bit = self.render_node(node, context)

  File "/opt/erigones/envs/lib/python2.7/site-packages/django/template/base.py", line 854, in render_node
    return node.render(context)

[...]

PR #17, opened 2016-12-12
Fixed issue #14

ricco386:


api/accounts/urls.py
```diff
@@ -12,15 +12,15 @@
     # /accounts/user/profile - get
     url(r'^user/profile/$', 'userprofile_list', name='api_userprofile_list'),
     # /accounts/user/<username> - get, create, delete
-    url(r'^user/(?P<username>[A-Za-z0-9\@\._-]+)/$', 'user_manage', name='api_user_manage'),
+    url(r'^user/(?P<username>[A-Za-z0-9@.+_-]+)/$', 'user_manage', name='api_user_manage'),
```
mbag: These regexes are too broad, they will match what we need, but also a lot that we might not like. However when signing up we are guarded by HTML5 email validation in our signup form. @dn0 what do you think?

gui/accounts/forms.py
```diff
@@ -201,6 +202,10 @@ def clean_email(self, *args, **kwargs):
             if User.objects.filter(Q(email__iexact=email) | Q(username__iexact=email)):
                 raise forms.ValidationError(_('This email address is already in use. Please supply a different email '
                                               'address.'))
+
+        if re.match("^[A-Za-z0-9@.+_-]+$", email) is None:
```
dn0: Please move this regexp to a class-level attribute with `re.compile`

gui/accounts/forms.py
```diff
@@ -201,6 +202,12 @@ def clean_email(self, *args, **kwargs):
             if User.objects.filter(Q(email__iexact=email) | Q(username__iexact=email)):
                 raise forms.ValidationError(_('This email address is already in use. Please supply a different email '
                                               'address.'))
+
+        regex = re.compile("^[A-Za-z0-9@.+_-]+$")
```
dn0: Attribute name `regex` is not very specific.

PR #19, opened 2016-12-14
16 hostname uuid resolution

### LINKED ISSUE #16: Reveal UUID of VM in API and GUI
* add support for working with VMs with uuid as required parameter (API)
* show uuid in GET vm_manage (`/vm/(hostname)`) and GET vm_define (`/vm/(hostname)/define`) output (read-only)
* show uuid in VM details (GUI)

secult:
Closes #16

### DIFF
--- a/api/vm/utils.py
+++ b/api/vm/utils.py
@@ -10,10 +10,10 @@ from vms.models import Storage, NodeStorage, Node, Vm, VmTemplate, Image, Subnet
 QNodeNullOrLicensed = Q(node__isnull=True) | ~Q(node__status=Node.UNLICENSED)
 
 
-def get_vm(request, hostname, attrs=None, where=None, exists_ok=False, noexists_fail=False, sr=('node',), api=True,
+def get_vm(request, hostname_or_uuid, attrs=None, where=None, exists_ok=False, noexists_fail=False, sr=('node',), api=True,
            extra=None, check_node_status=('POST', 'PUT', 'DELETE')):
     """
-    Call get_object for Vm model identified by hostname or uuid. If attributes are not
+    Call get_object for Vm model identified by hostname_or_uuid or uuid. If attributes are not
     specified then set them to check owner and node status.
     Also acts as IsVmOwner permission.
     """
@@ -31,7 +31,7 @@ def get_vm(request, hostname, attrs=None, where=None, exists_ok=False, noexists_
     if not request.user.is_admin(request):
         attrs['owner'] = request.user
 
-    attrs['hostname'] = hostname
+    attrs['hostname'] = hostname_or_uuid
     attrs['dc'] = request.dc
     attrs['slavevm__isnull'] = True
 
@@ -40,7 +40,7 @@ def get_vm(request, hostname, attrs=None, where=None, exists_ok=False, noexists_
     except (ObjectNotFound, ObjectAlreadyExists, Vm.DoesNotExist) as original_exception:
         # Checking whether user is not using uuid instead of the hostname, if not, throwing the original exception
         del attrs['hostname']
-        attrs['uuid'] = hostname
+        attrs['uuid'] = hostname_or_uuid
         try:
             vm = _get_vm_from_db(api, attrs, exists_ok, extra, noexists_fail, request, sr, where)
         except Exception:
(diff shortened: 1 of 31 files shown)


api/vm/utils.py
```diff
@@ -31,23 +51,20 @@ def get_vm(request, hostname, attrs=None, where=None, exists_ok=False, noexists_
...
-        if sr:
-            qs = Vm.objects.select_related(*sr)
-        else:
-            qs = Vm.objects
-
-        if where:
-            vm = qs.filter(where).get(**attrs)
-        else:
-            vm = qs.get(**attrs)
+    try:
```
dn0: Not sure about this block. It may lead to false results. Please re-think the logic

api/vm/utils.py
```diff
@@ -56,7 +56,18 @@ def get_vm(request, hostname_or_uuid, attrs=None, where=None, exists_ok=False, n
     attrs['dc'] = request.dc
     attrs['slavevm__isnull'] = True
 
-    vm = _get_vm_from_db(api, attrs, exists_ok, extra, noexists_fail, request, sr, where)
+    try:
+        vm = _get_vm_from_db(api, attrs, exists_ok, extra, noexists_fail, request, sr, where)
+    except (ObjectNotFound, Vm.DoesNotExist) as original_exception:
```
secult: How about now?

api/vm/utils.py
```diff
@@ -31,23 +52,22 @@ def get_vm(request, hostname, attrs=None, where=None, exists_ok=False, noexists_
...
-            vm = qs.filter(where).get(**attrs)
-        else:
-            vm = qs.get(**attrs)
+    try:
+        vm = _get_vm_from_db(api, attrs, exists_ok, extra, noexists_fail, request, sr, where)
+    except (ObjectNotFound, Vm.DoesNotExist) as original_exception:
+        del attrs['hostname']
+        attrs['uuid'] = hostname_or_uuid
+        try:
+            vm = _get_vm_from_db(api, attrs, exists_ok, extra, noexists_fail, request, sr, where)
```
dn0: Does it make sense to hit the DB twice for every non-existent VM? Wouldn't it be better to check if the provided argument is actually a uuid before asking the DB for the 2nd time? (we have `que.utils.validate_uuid()`, but you may also want to use a regexp since we store uuid's in only one format in the DB - write own validator if thats faster (api/validators.py).

api/vm/utils.py
```diff
@@ -10,10 +10,31 @@
 QNodeNullOrLicensed = Q(node__isnull=True) | ~Q(node__status=Node.UNLICENSED)
 
 
-def get_vm(request, hostname, attrs=None, where=None, exists_ok=False, noexists_fail=False, sr=('node',), api=True,
-           extra=None, check_node_status=('POST', 'PUT', 'DELETE')):
+def _get_vm_from_db(api, attrs, exists_ok, extra, noexists_fail, request, sr, where):
```
secult: They are ordered by alphabet but I will try to reduce them and put request back in front.

api/vm/utils.py
secult: By faster you mean faster for me or faster for the code execution? Also do you think this edge-edge case(user massively providing a nonexistent uuid in a wrong format) would have any relevant impact on performance?

dn0: I've meant code execution (including DB access).

OK, let me give you another view. If you would try out to create a VM with a hostname equal to some UUID you would see that is not possible. The uuid's are stored as strings in a specific format in the DB - they can be definitely tested either by `que.utils.validate_uuid()` or by a regex. If you look at the regexp for hostname in `api.vm.define.serializers.py` you can be pretty sure that a UUID cannot be passed as a hostname.
This means that you can simplify the code to a `if is_uuid(hostname)` and you should end up without needing the `_get_vm_from_db()` function.

secult: I used the validate_uuid function

dn0: If I'm not mistaken you should end up in a situation without any try/except block and without the helper function if you use the uuid check beforehand...

api/vm/utils.py
secult: We talked about backward compatibility - the precedence of the resolution by hostname should be done first. In case that hostname==uuid, this code would result in an error, but it should result in resolution of the correct object by hostname.

dn0:
This PR is OK to merge, but must be merged into v2.4 branch.

Issue #23, opened 2016-12-21
Add vm_uuid into output of VM-related API calls


dn0:
Continuation of issue #16 
The field should be a read-only field similar to the `hostname` attribute.
Affected files identified so far:
- api/vm/status/serializers.py
- api/vm/backup/serializers.py
- api/vm/snapshot/serializers.py
- api/mon/vm/serializers.py
- api/network/ip/serializers.py - this needs to be changed in issue #3 - which is already discussed in PR https://github.com/erigones/esdc-ce/pull/4#discussion_r92929413

PR #4, opened 2016-11-17
Added complete support for allowed_ips - issue #3

dn0:
Implemented via M2N IPAddress.vms field, which lists additional VMs
    associated with IP addresses that use the allowed_ips attribute.

PR #27, opened 2016-12-24
allowed_ips - multiple IPs per NIC

dn0:
Continuation of PR #4 with target branch changed to `v2.4`.
I've also updated to code to be compatible with changes stated in #23.


mbag:
I have problem understanding some of the decisions in code, namely when code branches with if-else constructs. It's not clear what is being accomplished with these "if" checks.  
Newly added help functions and variables (especially if they are defined far away from where they are used) need docstrings and comments to explain what purpose do they serve.

api/vm/define/serializers.py
```diff
@@ -1397,10 +1466,52 @@ def default_ptr(server, ip_address):
...
+                        ip_old, vm)
+        else:  # DB only operation
+            if many:
+                ip_old.vms.remove(vm)
+            else:
+                ip_old.vm = None
+                ip_old.save()
+
+    @staticmethod
+    def _update_vm_ip_association(vm, ip, delete=False, many=False):
```
mbag: split into 2 functions, the unfortunate verb `update` in the name of function suggests that something existing is being updated, however, when I look into the code it's either deleted or created. Please consider splitting into 2 functions, since delete and create don't don't share any code. IMHO this will yield more explicit design and 2 simpler functions.
Can we loose argument `many` and instead use `vm` argument to extract information about number of VMs that are being passed? It seems weird to pass additional info, when this info can be inferred/calculated from already passed argument.

api/vm/define/serializers.py
```diff
@@ -1447,6 +1535,15 @@ def save_ip(self, task_id, delete=False, update=False):
             if ip and self._net and self._net.ptr_domain:
                 self.save_ptr(self.request, task_id, vm, ip, self._net, delete=delete)  # fails silently
 
+        if update and not allowed_ips_old:
```
mbag: To get rid of the `pass` use `if not update or allowed_ips_old`, unless you plan to replace `pass` in the future with some business logic. In that case please document why are you using pass.

vms/models/subnet.py
```diff
@@ -127,7 +128,7 @@ def is_used_by_vms(self, dc=None):
                 return True
             vms = dc.vm_set.all()
         else:
-            if self.ipaddress_set.filter(vm__isnull=False).exists():  # Quick check
+            if self.ipaddress_set.filter(Q(vm__isnull=False) | ~Q(vms=None)).exists():  # Quick check
```
mbag: Quick check of what?

api/vm/define/serializers.py
dn0: Thanks. I've also changed a similar condition around line 1520.

vms/models/subnet.py
dn0: It is a quick check whether the IP address is used by some virtual machines (as the function name implies).  Nonetheless, I've added a comment block at the begging of the function, which should explain the code block.

api/vm/define/serializers.py
dn0: I don't agree with both comments :)
This function really updates the relationship between IP address and VM. It does not delete or creates any of those objects. I agree that if it would create or delete something than the name would be misleading. Also splitting the function into two would require another function or at least an `if` statement, because of the `delete` parameter (see how it is called)

We cannot loose the `many` parameter - it indicates whether to update the `ip.vm` ForeignKey or the `ip.vms` ManyToManyField. There is always in `ip` and one `vm` passed to the function.

api/vm/define/serializers.py
```diff
@@ -1185,6 +1194,30 @@ def validate_net(self, attrs, source):
 
         return attrs
 
+    def _is_ip_used(self, ipaddress, allowed_ips=False):
```
dn0: OK, so you see the inverted logic here - the function returns a specific error when the IP is used and `None` when it's not used. Any suggestion for a better name?

api/network/base/serializers.py
```diff
@@ -156,11 +156,19 @@ def update_errors(self, fields, err_msg):
...
+    _free_ip_subquery = '"vms_ipaddress"."vm_id" IS NULL AND "vms_ipaddress"."usage" = %d '\
+                        'AND "vms_ipaddress_vms"."vm_id" IS NULL' % IPAddress.VM
+
+    ips_free_query = 'SELECT COUNT(*) FROM "vms_ipaddress" LEFT OUTER JOIN '\
+                     '"vms_ipaddress_vms" ON ("vms_ipaddress"."id" = "vms_ipaddress_vms"."ipaddress_id") ' \
+                     'WHERE "vms_subnet"."uuid" = "vms_ipaddress"."subnet_id" '\
                      'AND %s' % _free_ip_subquery
-    ips_used_query = 'SELECT COUNT(*) FROM "vms_ipaddress" WHERE "vms_subnet"."uuid" = "vms_ipaddress"."subnet_id" '\
+
+    ips_used_query = 'SELECT COUNT(*) FROM "vms_ipaddress" LEFT OUTER JOIN '\
```
secult: Why is this in plain SQL?

secult: Maybe we should refactor these and similar in a separate task, not in this one. This is related to introduction of https://docs.djangoproject.com/en/1.10/releases/1.8/#query-expressions-conditional-expressions-and-database-functions in Django 1.8

Issue #24, opened 2016-12-21
Reveal snapshot and backup IDs


dn0:
Lets show the user actual snapshot names and backup paths in the API. The idea of this change is to enable users to easily identify certain snapshots/backups on the FS, which might be helpful in some (disaster) situations.

Suggested read-only attributes (let's discuss this first):

- Snapshot: `id` - `Snapshot.zfs_name`
- Backup: `file_path` - `Backup.file_path`

(This is a really simple change - 2 lines of code)

ricco386:
PR #37 closes this issue.

PR #36, opened 2017-01-04
Updated serializers to list new fields in API calls due to issue #24

ricco386:


ricco386:
Create PR to master, should go to branch v2.4

PR #37, opened 2017-01-04
Updated serializers to list new fields in API calls due to issue #24

ricco386:


api/vm/snapshot/serializers.py
```diff
@@ -130,3 +130,4 @@ class SnapshotSerializer(s.InstanceSerializer):
     created = s.DateTimeField(read_only=True, required=False)
     status = s.IntegerChoiceField(choices=Snapshot.STATUS, read_only=True, required=False)
     size = s.IntegerField(read_only=True)
+    zfs_name = s.SafeCharField(read_only=True)
```
mbag: I talked to @dn0 and we agreed that we need to talk about naming it differently.

PR #41, opened 2017-01-06
Added task retries on mgmt callbacks - closes #38

### LINKED ISSUE #38: Add task retry to mgmt callbacks
When a mgmt callback fails during execution because of an operational error (DBError, Redis/RabbitMQ connection problem) we should retry the task.

dn0:
When a mgmt callback fails during execution because of an operational error (DBError, Redis/RabbitMQ connection problem) the task will be retried (forever) using celery's task.retry() method.

Additional changes:

 - Celery has retry policy for result backends (redis), but it is hardcoded =>
     => a patch was added that changes the number of redis reconnection attempts to infinite.

 - The callback decorator has now access to the celery task context (request) (task's bind=True)

 - Now ignoring node status changes in DEBUG mode.

 - Fixed que.utils.delete_task() error when no workers are connected.


api/node/sysinfo/tasks.py
```diff
@@ -11,14 +11,15 @@
...
+from que.mgmt import MgmtCallbackTask
 from que.exceptions import TaskException
 
 __all__ = ('node_sysinfo_cb',)
 
 logger = get_task_logger(__name__)
 
 
-@cq.task(name='api.node.sysinfo.tasks.node_sysinfo_cb', ignore_result=False)
+@cq.task(name='api.node.sysinfo.tasks.node_sysinfo_cb', base=MgmtCallbackTask, bind=True, ignore_result=False)
```
secult: mgmt_lock decorator expects task_id as the first argument

dn0: Nice catch. Fixed in commit 4fe379f

PR #52, opened 2017-01-14
Make image server optional

dn0:
After this change the image server will be optional. It should be possible to configure other image server (or disable image server support) by adjusting the `VMS_IMAGE_VM` global setting. Additionally the user is able to configure other image sources on compute nodes - this will be useful when image server support is disabled.

When image server support is disabled, all /image/(name) API/GUI
functionality is performed only in the DB (synchronously).

New global settings:
 - VMS_IMAGE_VM - Global image server (hostname or uuid) (stored as uuid)
 - VMS_IMAGE_SOURCES - Additional IMGAPI sources

imgadm sources are now automatically updated on all compute nodes after
VMS_IMAGE_VM or VMS_IMAGE_SOURCES settings have changed. Also the imgadm
sources update mechanism has changed - now the `/var/imgadm/imgadm.conf`
file is being updated instead of calling imgadm.

Additionally a bug in the image update (`PUT /image/(name)`) was fixed
by introducing a Image.backup property. This property is used to restore
image attributes when an image update fails.

Internally the ImageVm was changed so it does not cache the related Vm object.


dn0:
Related commit in https://github.com/erigones/esdc-docs/commit/c8ec33476d27a93df0d9bf19df2c4ceca142d704

api/vm/utils.py
```diff
@@ -71,7 +73,7 @@ def get_vm(request, hostname_or_uuid, attrs=None, where=None, exists_ok=False, n
     return vm
 
 
-def get_vms(request, where=None, sr=('node',), order_by=('hostname',)):
+def get_vms(request, where=None, sr=('node',), order_by=('hostname',), dc_bound=True):
     """
     Return queryset of VMs for current user or all if admin.
```
dn0: I've added a comment explaining the dc_bound parameter

api/node/image/tasks.py
```diff
@@ -58,24 +59,48 @@ def node_image_cb(result, task_id, nodestorage_id=None, zpool=None, img_uuid=Non
...
+    Called by
     """
-    add_rem = ImageVm().sources_update(imgadm_sources)
+    if new_img_sources is None:
+        image_vm = ImageVm()
+        new_img_sources = image_vm.sources
 
-    if not add_rem:
-        logger.debug('Image sources already synced for node %s - skipping update', node)
+    if node_img_sources is not None and new_img_sources == node_img_sources:
```
dn0: Yes.

PR #56, opened 2017-01-20
Added explanations to DC settings GUI section

mbag:
The idea was to make editing of datacenter settings more intuitive.
The user should be able to figure out which settings can be edited
from which datacenter, since some of them are only editable from
default main datacenter.


mbag:
Requested changes were implemented.
Also rst note was added to dc_settings API docstring explaining that the global settings can only be changed form the main data center.

mbag:
@YanChii the typo in variable name was fixed. if there are no further objections I'll merge it into v2.4 branch.

PR #76, opened 2017-02-24
Added ``cpu_type`` parameter into vm_define API call

dn0:
The cpu_type needs to be set to ``host`` for Windows 2016 to work
correctly. For all other VMs the default value will be ``qemu64`` as it
always was.


api/vm/define/views.py
```diff
@@ -180,6 +183,8 @@ def vm_define(request, hostname_or_uuid, data=None):
         :type data.cpu_shares: integer
         :arg data.zfs_io_priority: IO throttle priority relative to other VMs (requires |SuperAdmin| permission)
         :type data.zfs_io_priority: integer
+        :arg data.cpu_type: Type of the virtual CPU exposed to the VM. One of qemu64, host (KVM only)
```
mbag: I think that *(KVM only)* should be at the beginning, because it relates to data.cpu_type. When I see like this, at the end, after the `host` I associate it with `host` i.e. that `host` type should be only used on KVM VMs. Furthermore, enclose `qemu64` and `host` in some kind of brackets or curly braces, so that it's clearer to user, what are their options.

api/vm/define/serializers.py
```diff
@@ -497,14 +499,19 @@ def validate_node_resources(self, attrs):
     def validate(self, attrs):
         vm = self.object
 
+        try:
+            ostype = attrs['ostype']
+        except KeyError:
+            ostype = vm.ostype
```
dn0: No, because in that case `vm.ostype` is executed regardless of existence of `attrs['ostype']`.

PR #78, opened 2017-02-28
Removed Terms of Service acceptation checkbox when empty TOS_LINK setting

dn0:
The TOS checkbox does not make sense without a valid HTTP link to Terms of Service.

### DIFF
--- a/gui/models/userprofile.py
+++ b/gui/models/userprofile.py
@@ -148,6 +148,7 @@ class UserProfile(models.Model):
                 (self.tos_acceptation or not settings.TOS_LINK) and
                 self.email_verified and
                 self.phone_verified and
+                # A company user account also requires a valid company ID
                 (self.usertype == self.PERSONAL or (self.usertype == self.COMPANY and self.companyid))
             )
         )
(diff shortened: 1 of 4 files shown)


gui/profile/forms.py
```diff
@@ -250,7 +250,8 @@ def __init__(self, request, profile, *args, **kwargs):
         super(UserProfileForm, self).__init__(request, profile, *args, **kwargs)
 
         # Display TOS acceptation checkbox if user did not accepted TOS and registration is enabled
```
mbag: `did not accept`  
Also please make sure that this comment describes code below and not it's negation. If I understand code correctly, the tos_acceptation is removed from self.fields dictionary i.e. not displayed when one of the conditions is met. This puts unnecessary strain on the readers brain. Furthermore, not all conditions are described by comment.

gui/models/userprofile.py
```diff
@@ -145,7 +145,7 @@ def is_ok(self):
         return (
             self.user.email and self.phone and (
                 not settings.REGISTRATION_ENABLED or
-                self.tos_acceptation and
+                (self.tos_acceptation or not settings.TOS_LINK) and
                 self.email_verified and
                 self.phone_verified and
                 (self.usertype == self.PERSONAL or (self.usertype == self.COMPANY and self.companyid))
```
mbag: I'm not sure that I understand this whole logical expression. Can we maybe break it into individual variables describing what each part means, e.g. personal_or_company_account = (self.usertype == self.PERSONAL or (self.usertype == self.COMPANY and self.companyid))

Issue #100, opened 2017-03-06
Node graphs not showing when not in main DC


mbag:
When adding nodem onitoring the call_mon_histoy_task() function was created. This new function is used by both Node and Vm to retrieve history. However, the dc_bound parameter, which for the node should be set to False.

PR #101, opened 2017-03-06
Fixed bug that caused node monitoring graphs not to show

mbag:
This only happened when user was not in main DC.
Issue #100

### DIFF
--- a/api/mon/utils.py
+++ b/api/mon/utils.py
@@ -75,7 +75,7 @@ def call_mon_history_task(request, task_function, view_fun_name, obj, dc_bound,
         tg = TG_DC_UNBOUND
 
     ter = task_function.call(request, obj.owner.id, (obj.uuid, items, history, result, items_search),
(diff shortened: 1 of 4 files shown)


api/mon/utils.py
```diff
@@ -66,10 +68,16 @@ def call_mon_history_task(request, task_function, view_fun_name, obj, serializer
         items_search = None
 
     history = graph_settings['history']
-    ter = task_function.call(request, obj.owner.id, (obj.uuid, items, history, result, items_search),
-                             meta={'apiview': _apiview_}, tidlock=tidlock)
+    # for VM the task_function is called without task group value because it's DC bound
+    if dc_bound:
+        ter = task_function.call(request, obj.owner.id, (obj.uuid, items, history, result, items_search),
+                                 meta={'apiview': _apiview_}, tidlock=tidlock)
```
dn0: Please make this one function call. Use the if only for setting a variable

PR #119, opened 2017-03-22
Various fixes to the Zabbix system on esdc-mon

dn0:
Related to erigones/esdc-factory#29, erigones/esdc-factory#30 and other commits in the 2.5.2 release of esdc-mon appliance.


ans/library/zabbix.py
```diff
@@ -0,0 +1,68 @@
...
+        module.fail_json(msg="Failed to connect to Zabbix server: %s" % e)
+        raise AssertionError
+
+    try:
+        result = zbx.call(api_method, api_params)
+    except ZabbixAPIException as e:
+        module.fail_json(msg="Zabbix API error: %s" % e)
+        raise AssertionError
+    except Exception as e:
+        module.fail_json(msg="Other failure: %s" % e)
```
YanChii: I prefer "Unknown failure"

dn0: @YanChii I've updated the message and also updated both library modules after running code inspection. I'll push the modules also to the esdc-mon role in https://github.com/erigones/esdc-factory.

PR #120, opened 2017-03-23
Decreased MON_ZABBIX_TIMEOUT to 15 seconds

dn0:
The default 60 seconds timeout was too high.

A long timeout is unnecessary, because the mgmt and mon VMs are on the same network. In case the mon VM is not reachable the mgmt would have to wait for 1 minute before displaying an error.


dn0:
@secult  I've updated the original PR description/comment. Thanks.

Issue #99, opened 2017-03-06
Update resource allocation for replicated VMs


dn0:
Add option to skip reservation of vCPUs and RAM for a replicated VM. These resources should be reserved on a compute node only if the VM is promoted to master there.

dn0:
Implemented in PR #122 and esdc-ee-replication

PR #122, opened 2017-03-25
Added SlaveVm.reserve_resources attribute - issue #99

dn0:
This attribute affects the reservation of CPU and RAM resources of the slave VM.
This will be used by the vm_replica API.
Documentation must be merged into esdc-docs after this PR is merged - erigones/esdc-docs#10


api/vm/utils.py
```diff
@@ -71,7 +71,7 @@ def get_vm(request, hostname_or_uuid, attrs=None, where=None, exists_ok=False, n
     return vm
 
 
-def get_vms(request, where=None, sr=('node',), order_by=('hostname',), dc_bound=True):
+def get_vms(request, where=None, sr=('node', 'slavevm'), order_by=('hostname',), dc_bound=True):
```
mbag: Please make sure, that changing default tuple in the function definition does not affect it usage on the following lines:
api/vm/backup/views.py at line 45
api/vm/snapshot/views.py at line 46

The get_vms() is used there with default values.

api/vm/define/views.py
```diff
@@ -36,7 +36,7 @@ def vm_define_list(request, data=None):
         :status 200: SUCCESS
         :status 403: Forbidden
     """
-    vms = get_vms(request, sr=('node', 'owner', 'template'), order_by=VmDefineView.get_order_by(data))
+    vms = get_vms(request, sr=('node', 'owner', 'template', 'slavevm'), order_by=VmDefineView.get_order_by(data))
```
mbag: Please make sure that this chang of the function definition defaults, does not affect usage on the following lines:
api/vm/backup/views.py at line 45
api/vm/snapshot/views.py at line 46

vms/models/vm.py
```diff
@@ -1226,10 +1230,17 @@ def get_disk_size(self, zpool=None):
 
         return dsk_size
 
-    def get_cpu_ram_disk(self, zpool=None, ram_overhead=False):
+    def get_cpu_ram_disk(self, zpool=None, ram_overhead=False, ignore_cpu_ram=None, ignore_disk=False):
```
mbag: Why is `ignore_cpu_ram` set to `None` instead of `False` as it is case with `ignore_dsik`. I would expect that 2 variables with the same name used for the similar thing, might use same data type.

api/vm/utils.py
dn0: The `'slavevm'` 1:1 relation should prefetched in any case. So this was intentional.

vms/models/vm.py
dn0: The 'slavevm' 1:1 relation should prefetched in any case. So this was intentional.

api/vm/define/views.py
dn0: The 'slavevm' 1:1 relation should prefetched in any case. So this was intentional.

PR #123, opened 2017-03-26
Updated set_request_method() and call_api_view() functions

dn0:
These functions are used whenever the code needs to call another view.
Usually, the `call_api_view()` function is used when performing internal API calls or by every call made via socket.io. The `set_request_method()` is sometimes needed when the caller does not want to call the whole view function, but only a smaller part of the view.

In both cases, the call to the new view should use a copy of the request object. The request object in the current function should not be modified by `set_request_method()` or `call_api_view()`. I've changed the `call_api_view()` so it always uses a copy of the request object. The `set_request_method()` also creates a copy of the request object (except when called from `call_api_view()`) and therefore there is no need to set the old method back.

The race condition was observed during `DELETE node_image_list`. The `NodeImageView.cleanup()` method calls `call_api_view()` multiple times in a for loop and every function call uses one request object. When the API view is triggered from GUI, the very same request object can be altered in an another call made from socket.io. I believe that this not the only use case and many more race conditions can be avoided with this simple fix.


api/utils/request.py
```diff
@@ -13,17 +15,20 @@ def is_request(obj):
     return isinstance(obj, REQUEST_CLASSES) and hasattr(obj, 'dc')
 
 
-def set_request_method(request, method):
+def set_request_method(request, method, copy_request=True):
     """
-    Django HttpRequest.method is mutable, but RestFramework Request.method is not.
+    This helper function changes the request.method attribute. By default, it creates a copy of the request object
+    and the caller should use the new copy. Callers who already have a copy of the request object will use the
+    copy_request=False parameter (call_api_view() and get_dummy_request()).
```
dn0: A second copy will be created (which is not bad).

api/utils/views.py
```diff
@@ -10,8 +11,11 @@ def call_api_view(request, method, fun, *args, **kwargs):
     """
     A handy wrapper for calling api view functions or classes.
     """
+    # We need a shallow copy of the request object because the original request object is still used by our caller
+    request = copy(request)
```
dn0: Since we are changing only the `request.method` attribute it should be OK. We treat the other request attributes as read-only variables.

Issue #96, opened 2017-03-06
Fix blank page on 403 (Forbidden)


dn0:

mbag:
@dn0 
We have two options here, currently we are returning `HttpResponseForbidden()` when there is permission problem with user's request. The solution is to render 403 template, that will be also added as part of the fix. The occurrences of HttpResponseForbidden will be substituted with:
`return HttpResponseForbidden(render(request, '403.html'))`

Another option is to substitute `return HttpResponseForbidden()` with `raise PermissionDenied` and let Django handle the template (I've tried it, the correct custom template is found and returned in the response).

PR #126, opened 2017-03-26
Correct handling of the 403 Forbidden

mbag:
New 403 template was added that is displayed if user tries to access resource that they are not supposed to view.

Issue #96

### DIFF
--- a/gui/templates/403.html
+++ b/gui/templates/403.html
@@ -1,0 +1,68 @@
+{% load i18n %}<!DOCTYPE html>
+<html>
+<head>
+  <title>{% trans "You are not permitted to view this page" %} (403)</title>
+  <style type="text/css">
+      body {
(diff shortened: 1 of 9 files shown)


gui/templates/403.html
```diff
@@ -0,0 +1,68 @@
...
+        background-attachment: fixed;
+        overflow: hidden;
+      }
+
+      .wrapper {
+        position: absolute;
+        top:50%;
+        margin-top: -85px;
+        left: 50%;
+        margin-left: -160px;
```
dn0: Currently, the wrapper is not centered. The correct value should be `-210px`. Please change that also in 404.html.

mbag:
I have also made changes to esdc-ee-replication since this was only esdc-ee-* repository that had call to HttpResponseForbidden.

PR #140, opened 2017-04-01
Added ability to import images from local image server

dn0:
### Updated imagestore API

Added ability to import images from local image server
    
This functionality can be used to import manually created images (i.e. images converted from other virtualization platforms), which were placed onto the image server directly on the compute node.

Documentation must be merged into esdc-docs after this PR is merged - erigones/esdc-docs#12

### Updated bin/esimg command

Added `list` and `get` subcommands for viewing the contents of the datasets folder and getting image manifests. Also added a `-c` parameter to the update subcommand for re-checking checksums of already imported images.


dn0:
@YanChii suggested to place the local image repository button in image import view on the right side of the screen.

Issue #115, opened 2017-03-16
500 AttributeError: 'unicode' object has no attribute 'iteritems' when doing VM undo


dn0:
When doing *undo* of VM definition:

```
Traceback (most recent call last):
  File "/opt/erigones/api/views.py", line 484, in dispatch
    response = handler(request, *args, **kwargs)
  File "/opt/erigones/api/decorators.py", line 91, in handler
    return func(*args, **kwargs)
  File "/opt/erigones/api/decorators.py", line 212, in wrap
    return fun(request, *args, **kwargs)
  File "/opt/erigones/api/vm/define/views.py", line 647, in vm_define_revert
    return VmDefineRevertView(request).put(vm, data)
  File "/opt/erigones/api/vm/define/utils.py", line 23, in wrap
    return fun(view, vm, *args, **kwargs)
  File "/opt/erigones/envs/lib/python2.7/site-packages/django/utils/decorators.py", line 145, in inner
    return func(*args, **kwargs)
  File "/opt/erigones/api/vm/define/vm_define.py", line 330, in put
    self.nice_diff(vm_diff))
  File "/opt/erigones/api/vm/define/vm_define.py", line 284, in nice_diff
    store.append(__disk_nics_add_rem(stuff, idname, xid))
  File "/opt/erigones/api/vm/define/vm_define.py", line 267, in __disk_nics_add_rem
    add_rem = ['%s=%s' % (key, to_string(val)) for key, val in iteritems(items) if key != id_name]
  File "/opt/erigones/envs/lib/python2.7/site-packages/django/utils/six.py", line 599, in iteritems
    return d.iteritems(**kw)
AttributeError: 'unicode' object has no attribute 'iteritems'
```

PR #144, opened 2017-04-03
Fixed allowed_ips errors when undo change on vm nic

ricco386:
We were receiving 500 error when allowed_ips is defined and user hits
undo button (issue #115). Once _diff_lists were aplied we received allowed_ips in add
section instead of change. I have changed serialiezr to have set instead
of list, which resulted in correct diff and also sanitaze if user adds
twice the same IP.

### DIFF
--- a/api/vm/define/serializers.py
+++ b/api/vm/define/serializers.py
@@ -943,7 +943,7 @@ class VmDefineNicSerializer(s.Serializer):
     allow_mac_spoofing = s.BooleanField(default=False)
     allow_restricted_traffic = s.BooleanField(default=False)
     allow_unfiltered_promisc = s.BooleanField(default=False)
-    allowed_ips = s.IPAddressArrayField(default=[], max_items=NIC_ALLOWED_IPS_MAX)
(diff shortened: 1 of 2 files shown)


api/vm/define/serializers.py
```diff
@@ -1114,7 +1115,7 @@ def detail_dict(self, **kwargs):
             ret['ip'] = self.object.get('ip', None)
             ret['netmask'] = self.object.get('netmask', None)
             ret['gateway'] = self.object.get('gateway', None)
-            ret['allowed_ips'] = self.object.get('allowed_ips', [])
+            ret['allowed_ips'] = self.object.get('allowed_ips', set())
```
dn0: This makes no sense. The user should see the allowed_ips as a list (json array).

api/vm/define/serializers.py
```diff
@@ -1277,7 +1278,7 @@ def validate(self, attrs):
                 self._ip_old = self._ip
                 self._ip = None
 
-        allowed_ips = attrs.get('allowed_ips', [])
+        allowed_ips = attrs.get('allowed_ips', set())
```
dn0: This makes no sense because it is converted to a set below.

PR #149, opened 2017-04-06
Implemented SOA serial incrementation

mbag:
Everytime the DNS record is updated the SOA will be incremented by one.


mbag:
@dn0  please check code again with PyCahrm, I've run flake8 and I don't think that there are any unused imports left.

PR #152, opened 2017-04-16
Do not display *pending* status when the target VM status was already reached

dn0:
## Updated `vm.status_display()` to not display *pending* status when the target VM status was already reached
    
There is a race condition visible in the GUI:
    
- start a stopped VM
- after the VM status changes to running switch to the console or snapshot view
- the VM may be now in pending state (blue icon in the VM menu)
    
This can happen because a new VM status is delivered from the compute node to the mgmt before the vm_status callback is finished. And we are returning a *pending* VM status when a `PUT vm_status` task is running.
    
Added a useful `vm.is_changing_status_to()` that returns a target VM status when `PUT vm_status` API function is running. The target status is deducted from the `action` attribute in the apiview dict. This function can help us to determine whether the VM has already reached its desired new status.

## Moved `vm.update_uptime()` closer to new VM status saving
    
It did not make sense to update the uptime and not saving it. This happened when a new VM state with an old change time propagated to `_save_vm_status()` function.


vms/models/vm.py
```diff
@@ -1268,6 +1274,21 @@ def is_changing_status(self):
         """Return appropriate task if VM status is changing"""
         return self.get_tasks(match_dict={'view': 'vm_status', 'method': 'PUT'})
 
+    def is_changing_status_to(self, vm_status_tasks=None):
```
mbag: The names of these two functions are rather unfortunate. The above function implies that it returns Bool, but in fact it returns task. 
And "is_changeging_status_to" function has almost the same name, but "_to" is appended at the end, which might not be obvious when reading code.

dn0: @mbag I agree. I've renamed the VM methods.

Issue #90, opened 2017-03-06
Add API and GUI for listing available monitoring templates and hostgroups


dn0:
The API call will be used for implementing templates suggestions dropdown in GUI.

dn0:
Feel free to modify the `get_task_result()` in `api.task.utils` to allow returning a list from an asynchronous API call.

PR #153, opened 2017-04-18
Implement monitoring template and hostgroup suggestions

mbag:
Issue #90


dn0:
@mbag I've have pushed an update with new formatting of the list comprehension.

I've also updated the updating of the select2 dropdown to be more user-friendly. However, I had to touch the select2.js code (one line).

mbag:
Checked out and applied new patch in the testing environment.
The code seems to work, if there are no other changes needed, the code can be merged into master.

PR #158, opened 2017-04-20
Updated PUT vm_manage to support forced updated of the node on VM

mbag:


api/vm/base/utils.py
```diff
@@ -103,3 +105,19 @@ def vm_reset(vm):
     cmd = 'vmadm stop %s -F; vmadm start %s' % (vm.uuid, vm.uuid)
     return execute(ERIGONES_TASK_USER, None, cmd, callback=False, queue=vm.node.fast_queue, nolog=True,
                    check_user_tasks=False)
+
+def vm_update(vm):
```
dn0: PEP8: two lines between function definitions.

api/vm/base/tasks.py
```diff
@@ -5,13 +5,15 @@
...
-from vms.models import Vm, Snapshot, SnapshotDefine, BackupDefine
-from vms.signals import vm_created, vm_deployed, vm_notcreated, vm_json_active_changed
+from vms.models import Vm, Snapshot, SnapshotDefine, BackupDefine, Node
+from vms.signals import vm_created, vm_deployed, vm_notcreated, vm_json_active_changed, vm_node_changed
 from api.email import sendmail
 from api.utils.request import get_dummy_request
 from api.task.tasks import task_log_cb_success, task_log_cb_error
 from api.task.utils import callback
-from api.vm.base.utils import is_vm_missing, vm_update_ipaddress_usage, vm_delete_snapshots_of_removed_disks, vm_reset
+from api.vm.base.utils import ( is_vm_missing, vm_update_ipaddress_usage,
```
dn0: PEP8: Whitespace after `(`.

api/vm/base/views.py
```diff
@@ -103,12 +103,18 @@ def vm_manage(request, hostname_or_uuid, data=None):
             * |async-yes|
         :arg hostname_or_uuid: **required** - Server hostname or uuid
         :type hostname_or_uuid: string
+        :arg data.force: If true, update VM definition according to current VM configuration on compute node
+        :type data.force: boolean
+        :arg data.node: (EXPERIMENTAL feature) Name of the compute node on which VM is located - \
+must be used used with `force` (requires |SuperAdmin| permission)
+        :type data.node: string
```
dn0: I suggest to add this warning to the `data.node` attribute:

```
.. warning:: The `node` parameter should be only used in special cases, i.e. disaster recovery from backups when the VM was restored manually on a different compute. This action does not perform any resource checking on the target compute node.

PR #176, opened 2017-06-05
add zookeepercli and query_cfgdb to cfgdb VM

YanChii:
CFGDB update to resolve erigones/esdc-factory#50


ans/upgrade/cfgdb/v2.6.0/main.yml
```diff
@@ -0,0 +1,32 @@
+# https://github.com/erigones/esdc-factory/issues/50
+- name: Download zookeepercli
```
dn0: Why don't we download it directly to cfgdb? (it should have the same network configuration as mgmt)

YanChii: I don't want to create new dependency on the internet during update of service virtuals. AFAIK, no mgmt update requires other service VMs to have internet connection.

PR #188, opened 2017-06-28
Fixed server error caused by invalid IP address value/s

mbag:
The problem was with input not being validated prior to
sending query to the database.

Please review this PR.
Thanks.

### DIFF
--- a/api/network/ip/api_views.py
+++ b/api/network/ip/api_views.py
@@ -54,7 +54,7 @@ class NetworkIPView(APIView):
                     for ip in ips:
                         ipaddress.ip_address(ip)
                 except ValueError:
-                    raise InvalidInput('Invalid IPs value')
+                    raise InvalidInput('Invalid ips value')
 
                 ip_filter.append(Q(ip__in=ips))
 
--- a/api/network/ip/api_views.py
+++ b/api/network/ip/api_views.py
@@ -1,4 +1,3 @@
-import ipaddress
 import operator
 from functools import reduce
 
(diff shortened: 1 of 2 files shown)


api/network/ip/api_views.py
```diff
@@ -48,6 +49,13 @@ def __init__(self, request, name, ip, data, dc=None, many=False):
             if ips is not None:
                 if not isinstance(ips, (tuple, list)):
                     raise InvalidInput('Invalid ips')
+                # test is IPs have valid format
```
secult: is->if typo

api/network/ip/api_views.py
```diff
@@ -48,6 +49,13 @@ def __init__(self, request, name, ip, data, dc=None, many=False):
...
+                try:
+                    for ip in ips:
+                        ipaddress.ip_address(ip)
+                except ValueError:
+                    raise InvalidInput('Invalid IPs value')
```
dn0: The property is called `ips` not `IPs`.

api/network/ip/api_views.py
```diff
@@ -48,6 +49,13 @@ def __init__(self, request, name, ip, data, dc=None, many=False):
             if ips is not None:
                 if not isinstance(ips, (tuple, list)):
                     raise InvalidInput('Invalid ips')
+                # test is IPs have valid format
+                try:
+                    for ip in ips:
+                        ipaddress.ip_address(ip)
```
dn0: Please use the static method `IPAddress.get_ip_address()`, which is the same, but in addition in does a type cast to unicode, which is important for this.

PR #192, opened 2017-07-03
Adjustable shutdown time for VMs

### LINKED ISSUE #111: Adjustable shutdown time for VMs
Currently, there's a global timeout when shutting down VMs. 3 minutes after sending an ACPI signal, the VM is forcefully shut down. Most of the time this is sufficient. But for special cases, it would be useful to have this timeout adjustable. E.g. on per-virtual-datacenter basis.

dn0:
Closes #111

Related PR in user guide: https://github.com/erigones/esdc-docs/pull/19


gui/static/gui/js/vm/vm-details.js
```diff
@@ -1110,13 +1118,18 @@ function vm_control(hostname, btn) {
 
     case 'stop':
     case 'reboot':
-      return vm_modal($('#vm_control_modal'), btn,
-        function() {
-          return vm_stop_or_reboot(hostname, action[1], false);
+      return vm_modal($('#vm_stop_or_reboot_modal'), btn,
+        function(e) {
+          if (validate_html_form($('#stop_timeout_period_form'))) {
```
dn0: This kind of code repeats too much. Any suggestion for improvement is welcome.

ricco386: Yes, it does. However, it always calls different function, that has different arguments.

In my opinion, if we abstract it, the abstract function would have too many input parameters which would result in code that is way too difficult to read and understand. This way anyone can who looks at it has an idea what it does, e.g. when you click on something (button `#stop_timeout_period_form`) it calls `vm_stop_or_reboot` or `vm_freeze` or something else and if I am not interested in the code below I have pretty good idea what is going on. In an abstract function, I would have to go there and inspect that is build and only afterward I would have an idea what is going on.

If there would be *args, **kwargs in javascript we could abstract if in a developer-friendly way, however, I am not aware of a way how to do it in JavaScript. So I think we should leave it as it is.

PR #194, opened 2017-07-04
Force vm status change in DB

### LINKED ISSUE #171: Add -force parameter to PUT /vm/(hostname)/status/current
This will be an administrative command to remove stale VM lock.

Only available to SuperAdmin.

mbag:
New method (PUT) for existing API call - closes #171

### DIFF
--- a/api/vm/status/views.py
+++ b/api/vm/status/views.py
@@ -78,6 +78,27 @@ def vm_status(request, hostname_or_uuid, action=None, data=None):
         :status 404: VM not found
         :status 423: Node is not operational / VM is not operational
 
+    .. http:put:: /vm/(hostname_or_uuid)/status/current
+
+        Retrieves current status from compute node.
+
+        :DC-bound?:
+            * |dc-yes|
+        :Permissions:
+            * |SuperAdmin|
+        :Asynchronous?:
+            * |async-yes|
+        :arg hostname_or_uuid: **required** - Server hostname or uuid
+        :type hostname_or_uuid: string
+        :arg data.force: Force change of the current status (default: false)
+        :type data.force: boolean
+        :status 200: SUCCESS
+        :status 201: PENDING
+        :status 400: FAILURE
+        :status 403: Forbidden
+        :status 404: VM not found
+        :status 423: Node is not operational
+
     .. http:put:: /vm/(hostname_or_uuid)/status/start
 
         :DC-bound?:
--- a/api/vm/status/views.py
+++ b/api/vm/status/views.py
@@ -61,7 +61,7 @@ def vm_status(request, hostname_or_uuid, action=None, data=None):
 
     .. http:get:: /vm/(hostname_or_uuid)/status/current
 
-        Retrieves current status from compute node.
+        Retrieves current VM status from compute node.
 
         :DC-bound?:
             * |dc-yes|
@@ -80,7 +80,7 @@ def vm_status(request, hostname_or_uuid, action=None, data=None):
 
     .. http:put:: /vm/(hostname_or_uuid)/status/current
 
-        Retrieves current status from compute node.
+        Retrieves current VM status from compute node and updates status value in DB for the VM.
 
         :DC-bound?:
             * |dc-yes|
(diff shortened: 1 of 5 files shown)


api/vm/status/vm_status.py
```diff
@@ -26,8 +26,8 @@ class VmStatus(APIView):
     """
     order_by_default = order_by_fields = ('hostname',)
     detail = ''
-    actions = ('start', 'stop', 'reboot')
-    statuses = (Vm.RUNNING, Vm.STOPPED, Vm.STOPPING, Vm.FROZEN)
+    actions = ('start', 'stop', 'reboot', 'current')
+    statuses = (Vm.RUNNING, Vm.STOPPED, Vm.STOPPING, Vm.FROZEN, Vm.NOTREADY_STOPPED, Vm.NOTREADY_RUNNING)
```
dn0: The `statuses` tuple is used in the original sense and therefore should not change.

api/vm/status/vm_status.py
```diff
@@ -146,16 +144,24 @@ def get(self, many=False):
             return SuccessTaskResponse(request, res)
 
         if self.action == 'current':
-            return self.get_current_status()
+            if vm.status not in (Vm.RUNNING, Vm.STOPPED, Vm.STOPPING, Vm.ERROR):
+                raise VmIsNotOperational
+            else:
```
dn0: Since the `if` statement acts as guard the `else` statement is not needed.

api/vm/status/vm_status.py
```diff
@@ -195,6 +201,11 @@ def put(self):
 
             return SuccessTaskResponse(request, res, task_id=tid, vm=vm)
 
+        elif action == 'current':
+            if not self.data.get('force', False):
+                raise ExpectationFailed('Force parameter must be used!')
```
dn0: Please `PreconditionRequired` exception and document it in `views.py`.

api/vm/status/vm_status.py
```diff
@@ -146,16 +144,24 @@ def get(self, many=False):
...
+            else:
+                return self.get_current_status()
         else:
             ser = VmStatusSerializer(vm)
             return SuccessTaskResponse(request, ser.data, vm=vm)
 
     def put(self):
         request, vm, action = self.request, self.vm, self.action
 
+        # for PUT /current/ action user needs to be SuperAdmin
```
dn0: In order to stay consistent with the rest of the code I'm suggesting to move this `if` guard into the `elif action == 'current'` block.

api/vm/status/views.py
```diff
@@ -78,6 +78,27 @@ def vm_status(request, hostname_or_uuid, action=None, data=None):
         :status 404: VM not found
         :status 423: Node is not operational / VM is not operational
 
+    .. http:put:: /vm/(hostname_or_uuid)/status/current
+
+        Retrieves current VM status from compute node and updates status value in DB for the VM.
+
```
dn0: Please add this warning message here:
```
    .. warning:: This API call is intended for internal administrative purposes only and should be used with care.
```

api/vm/status/vm_status.py
```diff
@@ -105,9 +105,7 @@ def _start_cmd(self, iso=None, iso2=None, once=False):
     def get_current_status(self):
         """Get current VM status"""
         request, vm = self.request, self.vm
-
-        if vm.status not in (Vm.RUNNING, Vm.STOPPED, Vm.STOPPING, Vm.ERROR):
-            raise VmIsNotOperational
+        force = self.data.get('force', False)
```
dn0: The value for `force_change` can be passed via a function parameter instead of getting it from data (it was already fetched in the caller).

api/vm/status/vm_status.py
mbag: Should I add this to other PUT actions as well? Since stop, reboot, and and start also raise PreconditionFailed, however none of them has 412 documented as output.

dn0: You mean 428...

mbag: Yes 428, it's only documented for action start but not the other actions.

PR #196, opened 2017-07-06
Raising ValueError when VM configuration bypassed our vm_define API

### LINKED ISSUE #168: IP addr database not refreshed after "es set /vm/<hostname> -force"
Hi,
when calling an update of VM configuration after manual changes using vmadm (es set /vm/<hostname> -force), the VM info is refreshed properly.. but the ESDC IP address database is not updated (considering there was an IP addr change). Therefore the new VM IP is treated as free and can be assigned to another host without a warning.
J.

dn0:
Closes #168


api/vm/base/utils.py
```diff
@@ -65,10 +75,40 @@ def vm_update_ipaddress_usage(vm):
...
+    for nic_id, nic in enumerate(vm.json_active_get_nics(), 1):
+        network_uuid = nic.get('network_uuid', None)
+
+        if network_uuid:
+            ip = nic.get('ip', '')
+            allowed_ips = nic.get('allowed_ips', ())
+
+            if ip:
+                if _is_ip_ok(vm_ips, ip, network_uuid):
+                    logger.debug('VM %s NIC ID %s IP address %s is OK', vm, nic_id, ip)
```
mbag: Seems unnecessary, to have inside of the if-block only for the purpose of logging debug statement. I would suggest using negation in the if-statement and raising ValueError, instead. Debug statement, if needed can be added outside this guard block.

api/vm/base/utils.py
```diff
@@ -65,10 +75,40 @@ def vm_update_ipaddress_usage(vm):
...
+                else:
+                    raise ValueError('VM %s NIC ID %s IP address %s is not properly associated with VM!' %
+                                     (vm, nic_id, ip))
+
+            for ip in allowed_ips:
+                if _is_ip_ok(vm_allowed_ips, ip, network_uuid):
+                    logger.debug('VM %s NIC ID %s allowed IP address %s is OK', vm, nic_id, ip)
```
mbag: Same as the comment above.

PR #197, opened 2017-07-06
Added VM update capability to VM reboot and stop operations

### LINKED ISSUE #170: Add update flag to vm_status reboot action
Similar to the `update` flag in vm_status start action.

dn0:
Closes #170


api/vm/base/vm_manage.py
```diff
@@ -81,6 +81,16 @@ def fix_update(json_update):
 
         return json_update, cmd
 
+    # noinspection PyUnusedLocal
+    @staticmethod
+    def validate_update(vm, json_update, os_cmd):
+        """Check if (json_update, os_cmd) tuple from fix_update() can be run on a VM"""
```
dn0: Because the update dict is the main product of `vm.json_update()` and `self.fix_update()` (which additionally produces an os_cmd string). It contains information about what is going to be updated on the node.

The purpose of the `validate_update()` is to check whether the update can be performed.
Currently we only check the `os_cmd` parameter, but I've a plan to move more validation logic into this function in the future.

mbag: ok

PR #211, opened 2017-07-26
Librabbitmq removal

### LINKED ISSUE #207: Memory leak in erigonesd on node with network down
On compute node: Cut connection from erigonesd to rabbitmq/redis and watch constant memory usage increasing.

secult:
Closes #207.

### DIFF
--- a/ans/upgrade/mgmt/v2.6.1/main.yml
+++ b/ans/upgrade/mgmt/v2.6.1/main.yml
@@ -1,0 +1,2 @@
+- name: Uninstall librabbitmq
+  shell: "{{ erigones_home }}/bin/ctl.sh pip_uninstall"
--- a/ans/upgrade/mgmt/v2.6.1/main.yml
+++ b/ans/upgrade/mgmt/v2.6.1/main.yml
@@ -1,2 +1,2 @@
 - name: Uninstall librabbitmq
-  shell: "{{ erigones_home }}/bin/ctl.sh pip_uninstall --package librabbitmq || true"
+  shell: "{{ erigones_home }}/bin/ctl.sh pip_uninstall --package librabbitmq --silence-errors"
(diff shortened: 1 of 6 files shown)


ans/upgrade/mgmt/v2.6.1/main.yml
```diff
@@ -0,0 +1,2 @@
+- name: Uninstall librabbitmq
+  shell: "{{ erigones_home }}/bin/ctl.sh pip_uninstall --library librabbitmq"
```
YanChii: Please add package version checking. What if we add another librabbitmq package in the future? It will remove it on accidental repeated run.
Same in node update.

core/management/commands/pip_uninstall.py
```diff
@@ -1,14 +1,17 @@
-from ._base import DanubeCloudCommand, lcd
+from ._base import DanubeCloudCommand, lcd, CommandOption
 
 
 class Command(DanubeCloudCommand):
-    help = 'Uninstall dependencies according to *etc/requirements-remove.txt*.'
+    help = 'Shortcut for pip uninstall within the application\'s virtual environment.'
+    option_list = (
+        CommandOption('--library', action='store', dest='library', default='',
```
dn0: Please change the `--library` to something more meaningful, e.g. `--package`, `--package-name`, ... And add support for single dash parameter (`-p`) (like other commands).

ans/upgrade/mgmt/v2.6.1/main.yml
secult: In case there will  be the same library installed again, the relevant script will take care of it. So for example, the library will be uninstalled in the 2.6.1 step and installed in 2.6.2 step.

secult:
Ansible script should ignore the error when a library does not exist (idempotency)

core/management/commands/pip_uninstall.py
```diff
@@ -1,14 +1,24 @@
...
+    def pip_uninstall(self, package, raise_on_error, params='-y'):
+        erroneous_return_code = self.local('pip uninstall %s %s' % (params, package), raise_on_error=raise_on_error)
+        if erroneous_return_code:
+            self.display('An error has occured while uninstalling %s!\n\n ' % package, color='red')
+            if not raise_on_error:
+                self.display('The library may be uninstalled already, we suppress the error.\n', color='magenta')
+        else:
+            self.display('%s have been successfully uninstalled.\n\n ' % package, color='green')
 
-    def handle(self, *args, **options):
```
dn0: The package should be a positional argument, because it is required.

Issue #98, opened 2017-03-06
User editable notes for VM and node (wiki?)


dn0:
The data (notes) can be stored in existing json fields in DB.

dn0:
Feature requested by [Alfa Centauri](http://www.alfac.sk/).

dn0:
Don't forget about templates if implemented as a new API call.

ricco386:
Functionality will be done in two phases.

The initial phase is going to 2.6.0 release and will contain new text DB fields for VM and node to store note. Updated functionality in API calls: vm/define and node/define to be able to store, show note. GUI has been amended to show this new fields via API calls. Currently, we support only plain text that is escaped.

The second phase will be in 2.7.0 release and will contain text field formatting in some reasonable format. We need to research some JavaScript library, that can format either ReStructured Text, Markdown, or some wiki engine and the user can store formatted text notes, that will be displayed nicely in the GUI (don't forget API validation!). Additionally, add a new screen that can display long text block, and note in details page should be truncated with a link to this new screen.

ricco386:
**The first phase has been done** and is part of 2.6.0. See https://github.com/erigones/esdc-ce/issues/98#issuecomment-313396929

ricco386:
**The second phase has been finished** and merged in PR #214

PR #214, opened 2017-07-31
Added support for markdown in vm and node notes field

ricco386:
This way we can escape html, and more importantly javascript in order
to prevent XSS, but have nice formatable notes.

Functionality added due to issue #98

### DIFF
--- a/etc/requirements-both.txt
+++ b/etc/requirements-both.txt
@@ -13,3 +13,4 @@ simplejson==3.11.1
 frozendict==1.2
 requests==2.13.0
 esdc-api==2.0.1
+Markdown==2.6.8
--- a/etc/requirements-both.txt
+++ b/etc/requirements-both.txt
@@ -13,4 +13,3 @@ simplejson==3.11.1
 frozendict==1.2
 requests==2.13.0
 esdc-api==2.0.1
-Markdown==2.6.8
--- a/gui/static/gui/js/node/node-details.js
+++ b/gui/static/gui/js/node/node-details.js
@@ -24,13 +24,11 @@ function Node(hostname) {
 
       mod.on('focus', '#id_note', function () {
         mod.off('keypress');
-        $(this).width('303px');
-        $(this)[0]['rows'] = 15;
+        $(this).addClass("large_modal_textarea");
       });
 
       mod.on('focusout', '#id_note', function () {
-        $(this).width('186px');
-        $(this)[0]['rows'] = 5;
+        $(this).removeClass("large_modal_textarea");
       });
     });
 
--- a/gui/static/gui/js/node/node-details.js
+++ b/gui/static/gui/js/node/node-details.js
@@ -19,7 +19,20 @@ function Node(hostname) {
   }
 
   $('#node_define').click(function() {
-    new obj_form_modal($(this), '#node_define_modal');
+    new obj_form_modal($(this), '#node_define_modal', function(mod) {
+      var mod = mod.mod;
+
+      mod.on('focus', '#id_note', function () {
+        mod.off('keypress');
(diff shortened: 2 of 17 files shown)


gui/templatetags/gui_utils.py
```diff
@@ -304,3 +305,8 @@ def qs_del(query_string, param):
     qs = QueryDict(query_string, mutable=True)
     qs.pop(param, None)
     return qs.urlencode()
+
+
+@register.filter
+def markdownify(text):
+    return markdown.markdown(text, safe_mode='escape')
```
dn0: Reading the documentation https://pythonhosted.org/Markdown/reference.html#markdown:
- `output_format` should be `html5`
- `safe_mode` is deprecated -> https://pythonhosted.org/Markdown/release-2.6.html#safe_mode-deprecated

gui/templatetags/gui_utils.py
```diff
@@ -7,6 +7,7 @@
 from json import dumps
 from datetime import datetime, timedelta
 
+import markdown
```
dn0: Please move the import statement under` import ipaddress`

gui/templatetags/gui_utils.py
ricco386: we dont mind safe_mode is deprecated as we will rely on django template escape, e.g. in template we escape the content than run via markdown and mark safe for django template to render.

etc/requirements-both.txt
```diff
@@ -13,3 +13,4 @@ simplejson==3.11.1
 frozendict==1.2
 requests==2.13.0
 esdc-api==2.0.1
+Markdown==2.6.8
```
secult: Are you sure that this should go to "requirements-both" and not to the "requirements-mgmt"?

gui/static/gui/js/node/node-details.js
```diff
@@ -19,7 +19,20 @@ function Node(hostname) {
   }
 
   $('#node_define').click(function() {
-    new obj_form_modal($(this), '#node_define_modal');
+    new obj_form_modal($(this), '#node_define_modal', function(mod) {
```
secult: When I click enter as a newline in the text, the modal closes.

gui/templatetags/gui_utils.py
```diff
@@ -304,3 +305,19 @@ def qs_del(query_string, param):
...
     qs.pop(param, None)
     return qs.urlencode()
+
+
+def _markdownify(text):
+    text = conditional_escape(text)
+
+    try:
+        md = markdown.markdown(text)
+    except:
```
secult: Too broad exception clause

gui/templatetags/gui_utils.py
```diff
@@ -304,3 +305,19 @@ def qs_del(query_string, param):
     qs = QueryDict(query_string, mutable=True)
     qs.pop(param, None)
     return qs.urlencode()
+
+
+def _markdownify(text):
+    text = conditional_escape(text)
+
+    try:
+        md = markdown.markdown(text)
```
secult: I would rename the md to something more meaningful

etc/requirements-both.txt
dn0: Definitely `requirements-mgmt.txt`

gui/templatetags/gui_utils.py
ricco386: We want too broad exception here, as we don't know what can get wrong in the markdown library and we don't want to have broken page in case text formatting in template tag got wrong...

secult: Even in that case, you want to use "Exception" class, not everything (including SystemExit).

PR #218, opened 2017-08-04
Changed redirect after user removes permissions to his/her current DC

### LINKED ISSUE #108: Behavior after user permission change is confusing
1. User is a UserAdmin in a datacenter that is not the main datacenter. 
2. User changes permission-related property on herself.
3. After clicking "Update", one gets immediately 403 Forbidden message.
(User was switched to _main_ datacenter and therefore cannot access _/dc*_ path. where she's not Admin).
The correct behavior may be that after permission related change, user should be redirected to url, where she has access (root url is sufficient).

dn0:
Closes #108

- commit ee6d06fea17234baaf17098ca1c106018d0abe79 : Changed redirect after user removes permissions to his/her current DC 
- commit c29c96bcc695fd938e77ca83f984c77b13dbd149 : Implemented direct task events, which are sent to only one specific user
- commit e41697a927ea4d5a3d96812ed51da83078133b68 : Added UserCurrentDcChanged direct event
- commit b98eba04a0d34d27ef5385b5a22f8a35271ef7ba : Updated changelog


gui/dc/user/views.py
```diff
@@ -97,7 +97,13 @@ def dc_user_modal_form(request):
                 return redirect('dc_user_profile', username=form.cleaned_data['username'])
             else:
                 messages.success(request, _('User profile was successfully updated'))
-                return redirect('dc_user_list', query_string=qs)
+                # You can modify yourself and lose access to /dc - Issue #108
+                if request.user == user and not user.is_admin(dc=request.dc):
```
secult: Think also about superadmin permission

gui/static/gui/js/gsio.js
```diff
@@ -721,6 +721,10 @@ function _task_event_callback(result) {
       alert2(result.message);
       break;
 
+    case 'user_current_dc_changed':
+      alert2(gettext('Your current datacenter was changed. Please refresh your browser.'));
```
secult: has

dn0: I like the passive form (meaning the DC was changed by someone else)

gui/dc/user/views.py
dn0: Fixed

dn0:
@secult checkout the updated code

secult:
I am not sure about the behaviour in this case:
Group A is attached To DC Third and contains User 1, User 2.
User 1 removes User 1 from the group. Both get message "Your DC was changed..." and User's 1 DC was changed, but User's 2 DC was not changed.

secult:
I have a problem with the following scenario:
User 1 has Admin access to DC B, but has no access to features of the default (main) DC. 
After changing permissions on DC B by somebody else,  user 1, which is in the /dc settings gets a message "Your datacenter was changed, please refresh your browser", but after pressing refresh he gets 403, because he doesn't have access to any of the /dc settings of the main datacenter and he did not get redirected anywhere as he would get in case he has permissions to that DC.

secult:
DC change should happen only after loss of permissions (check before revoking).

dn0:
@secult 

I've added 4 commits:

* commit 14d2fe0 Resetting of user's current_dc should not affect SuperAdmins
* commit f6626ce Fixed dc_manage to update current_dc of affected user's via user.current_dc property
* commit 32a0802 and commit a7fb754 Fixed POST/DELETE dc_group - **critical**
* commit 02e035f Changed message displayed to a user whose DC has changed

----

Regarding your last comment: Currently we reset user's current DC in a pessimistic way - we do this also on some occasions where the user's access to his current DC may remain, but we reset it to default anyway. In the commits above I've removed SuperAdmins from this logic. But I prefer to keep the code as it is, because of simplicity.

----
And the reason why all this is necessary is written in comments in the code and I will post it also here:

> When groups are changed on a non-default DC, we have to update the current_dc on every affected user, because she/he could remain access to this DC (this is because get_dc() uses current_dc as a shortcut)

> Plus, After changing DC groups we have to invalidate the list of admins for the affected DC

api/dc/base/dc_view.py
```diff
@@ -166,7 +166,8 @@ def put(self):
                 ser.owner_changed.current_dc = default_dc
 
             if ser.removed_users:
-                ser.removed_users.exclude(is_staff=True).update(default_dc=default_dc)
+                for user in ser.removed_users.select_related('default_dc').exclude(is_staff=True):
+                    user.current_dc = default_dc
```
secult: Shouldn't you _user.save()_ ? (also at the other places)

secult:
When DC owner changes, we don't show the message and don't change the DC. + Even though VM list is not visible anymore, task list is still visible, containing information about the Dc that the person just lost the permissions.

api/dc/base/dc_view.py
dn0: The `user.current_dc` is a setter property that does the `save()`, but I already though about replacing it with a classic method. What do you think?

dn0:
@secult I will look at the DC owner case deeper

dn0:
@secult OK, fixed in commit c131658 - originally we didn't save the old DC owner object

dn0:
@secult I've refactored changing of `user.current_dc` to go through a `user.reset_current_dc()` method - commit 054fc7f
This should improve readability (especially give the reader direct insight to the code's primary purpose).

gui/models/user.py
```diff
@@ -185,13 +191,14 @@ def current_dc(self, dc):
             self.save(update_fields=('default_dc',))
             # Send direct event to the user working in the old DC
             UserCurrentDcChanged(self.id, dc_id=old_current_dc_id).send()
+            return True
```
secult: We don't rely anywhere on the return value of this function

Issue #230, opened 2017-08-10
SMSAPI return response status code 200 but text of the response is ERROR


ricco386:
While working at issue #225 I have noticed strange behavior, I set use all the SMS_* settings and the default value SMS_SMSAPI_FROM has not been defined at SMSAPI. I made a registration and didnt receive any sms message, however as admin I didnt notice any error. Problem is that SMSAPI can return response status code 200 but text of the response is ERROR:

`2017-08-10 08:11:18,650 INFO [api.sms.views]: SMS has been sent to +421*********, status_code=200, response="ERROR:14"`

We need to add validation at SMSAPI response in order to be able to solve this specifically for this sms provider.

List of status codes and Error codes: https://www.smsapi.com/resp

PR #231, opened 2017-08-10
SMSAPI return response status code 200 but text of the response is ERROR

ricco386:
Fixed issue #230


api/sms/smsapi/views.py
```diff
@@ -47,6 +47,20 @@ def login_data():
     }
 
 
+def _smsapi_response_adapter(response):
+    """
+    Adapter to correct response code if there is error with the SMS.
+    SMSAPI return response status code 200 but text of the response is ERROR #230
```
dn0: return -> returns

api/sms/smsapi/views.py
```diff
@@ -47,6 +47,20 @@ def login_data():
...
+    """
+    if response.status_code == requests.codes.ok and response.text.startswith('ERROR'):
+        new_status_code = 406
+        logger.warning('Response got status code %s and contain "%s". Updating status code to %s!',
```
dn0: contain -> contains

dn0: `'Response got ...'` -> `'SMSAPI response got ...'`  # so we know that this about SMSAPI and not other response

PR #232, opened 2017-08-11
Renamed `GET /task/log/report` to `GET /task/log/stats`

dn0:
Refactored task_log_report to task_log_stats to be consistent with future stats API views. Plus, this commit fixes documentation issues and removes obsolete code.

### DIFF
--- a/api/task/task_log.py
+++ b/api/task/task_log.py
@@ -1,6 +1,8 @@
 from datetime import timedelta
 
+from celery import states
 from django.utils import timezone
+from django.db.models import Count
 
 from api import status
 from api.api_views import APIView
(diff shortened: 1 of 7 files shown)


api/task/task_log.py
```diff
@@ -38,14 +40,31 @@ def get(self):
...
-        report = TaskLogReportSerializer.get_report(qs)
+        res = self._get_stats_result(qs)
+        res['_last'] = last
+
+        return res
 
-        return TaskSuccessResponse(self.request, TaskLogReportSerializer(report).data)
+    def get_stats_response(self):
+        """api.task.views.task_log_stats"""
+        return TaskSuccessResponse(self.request, self.get_stats())
```
secult: What's the point of abstracting get_stats function when there is no usage of the function anywhere else, and it's also neither reducing complexity nor improving readability?

doc/api/source/api/task.rst
```diff
@@ -162,3 +162,30 @@
                 ]
             }
         }
+
+
+/task/log/stats
```
secult: I cannot manage to find this in the documentation (after gendoc)

PR #233, opened 2017-08-15
Added `GET /system/stats` API call

dn0:
The system_stats API call is a synchronous functions that returns various stats about datacenters, compute nodes and virtual servers in the whole system (from DB).


api/system/stats/api_views.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class SystemStatsView(APIView):
+    """api.system.stats.views.system_stats"""
+    _cache_key = 'system_stats'
+    dc_bound = False
+
+    @classmethod
+    def get_stats(cls, from_cache=True, cache_timeout=30):
+        if from_cache:
```
mbag: since method has from_cache param, I would expect to get cached data every time. However, this will not happen, if cache.get() fails. In that case, the method continues and returns non-cached data, which might not be what user wanted.

Issue #225, opened 2017-08-08
Simplify registration and password reset


dn0:
Users are not aware that after opening a unique link (registration/password reset) they need to click on the button to confirm the password reset.

secult:
I would suggest to use the default django password reset workflow

ricco386:
We use two auth registration and password rest, where we send password via sms message. This have to stay build in.

ricco386:
Current status:

Registratin steps:
* Click Sign in
* Fill the form and submit
* User see a message (email with verification link is send out)
* User click on email link
* User see a message (sms with password is send out)


Reset password steps:
* Click Forgotten password
* Fill the form and submit 
* User see a message (email with verification link is send out)
* User click on email link
* User see a message and click on confirm button
* User see a message (sms with password is send out)

ricco386:
After long discussion with @secult we have identified few possible improvements (not sure if any will be implemented, but I wanted to make a note). Possible issue: generated passwords vs. user created passwords.

Registration asks for phone and email to be sure that whoever registered is a real person. However password reset does not need this, we already know that user is real person. Therefore email should be enough. Other possibility is that after click on unique one time link user can enter password by himself. Since we dont want to lower security for existing installations (and change backwards functionality behavior), there shall be setting Either send password via email, or via phone (leave as it is now).

Possible problem is that user dosnt have the phone anymore his administrator have to change his phone number in order to be able to reset the password.

Instead of sending password via sms, we will just issue one time token via sms and user can enter his password (either through registration or thought forgotten password), and immediately after login he have to enter this one time token.

ricco386:
Created PR #234 Most of the changes are just a text changes and formated text into blocks. I have removed one step during the password reset (so the form is auto submitted).

Some issues discussed [above](https://github.com/erigones/esdc-ce/issues/225#issuecomment-321569755) has been created as a separate issue #235

dn0:
Merged PR #234 
Issue continues with #235

PR #234, opened 2017-08-15
Updated registration help texts in issue #225

ricco386:
I have simplified texts on the form (two lines of small words) was
barely readable, and added SMS in the text so user can notice.

Updated email texts to have explaining sentence and also moved the
data into "blocks" so it is more visualy more obvious to notice
stars insttead of password so user will notice that password will
be send out by text message (SMS).

Reset password form is automatically submitted by javascript in
order to send password via SMS. We do re-use the Django reset from
that is the reason we have to have the form there and submit.

### DIFF
--- a/gui/templates/gui/accounts/forgot_check.html
+++ b/gui/templates/gui/accounts/forgot_check.html
@@ -1,12 +1,14 @@
 {% extends "base_no_auth.html" %}
 {% load i18n %}
 
+{% block body_class %} onload="document.getElementById('submitForm').submit();"{% endblock %}
+
 {% block login_content %}{% if validlink %}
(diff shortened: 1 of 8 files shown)


gui/templates/gui/accounts/forgot_check.html
```diff
@@ -1,12 +1,14 @@
 {% extends "base_no_auth.html" %}
 {% load i18n %}
 
+{% block body_class %} onload="document.getElementById('submitForm').submit();"{% endblock %}
```
dn0: We don't use the HTML JS attributes, instead we typically use something like this:
```
<script type="text/javascript">
  $(document).ready(function() {
     ....
```

PR #243, opened 2017-08-22
made allowed ips type list everywhere

secult:
It was not possible to serialize (and therefore deploy) VM after using allowed_ips parameter in the VM definition, as the type of allowed_ips was set to set. Set is not JSON serializable. 
Default values were changed to promote the expected value in the variable.

### DIFF
--- a/api/node/vm/utils.py
+++ b/api/node/vm/utils.py
@@ -165,7 +165,7 @@ def vm_from_json(request, task_id, json, dc, owner=settings.ADMIN_USER, template
                 if i == 0:
                     primary_ip = ip
 
-                for ipaddress in nic.get('allowed_ips', ()):
+                for ipaddress in nic.get('allowed_ips', []):
                     _, err = _vm_save_ip_from_json(vm, net, ipaddress, allowed_ips=True)
 
                     if err:
--- a/api/vm/base/utils.py
+++ b/api/vm/base/utils.py
@@ -92,7 +92,7 @@ def vm_update_ipaddress_usage(vm):
 
         if network_uuid:
             ip = nic.get('ip', '')
-            allowed_ips = nic.get('allowed_ips', ())
+            allowed_ips = nic.get('allowed_ips', [])
 
             if ip:
                 logger.debug('VM: %s | NIC ID: %s | NIC network: %s | IP address: %s', vm, nic_id, network_uuid, ip)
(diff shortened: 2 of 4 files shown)


doc/changelog.rst
```diff
@@ -24,6 +24,7 @@ Features
 Bugs
 ----
 
+- Fixed allowed_ips type on all occurrences to list instead of set to enable JSON serialization - `#242 <https://github.com/erigones/esdc-ce/issues/242>`__
```
dn0: Wrong changelog entry!

api/vm/define/serializers.py
```diff
@@ -1059,11 +1059,11 @@ def fix_before(self, data):
             except IPAddress.DoesNotExist:
                 raise APIError(detail='Unknown ip in NIC definition.')
 
-        allowed_ips = data.get('allowed_ips', None)
+        allowed_ips = data.get('allowed_ips', [])
```
dn0: Are you sure about this? Directly below is a check for `None`...

api/node/vm/utils.py
```diff
@@ -165,7 +165,7 @@ def vm_from_json(request, task_id, json, dc, owner=settings.ADMIN_USER, template
                 if i == 0:
                     primary_ip = ip
 
-                for ipaddress in nic.get('allowed_ips', ()):
+                for ipaddress in nic.get('allowed_ips', []):
```
dn0: This makes no sense...

api/vm/base/utils.py
```diff
@@ -92,7 +92,7 @@ def vm_update_ipaddress_usage(vm):
 
         if network_uuid:
             ip = nic.get('ip', '')
-            allowed_ips = nic.get('allowed_ips', ())
+            allowed_ips = nic.get('allowed_ips', [])
```
dn0: Why?

api/vm/define/serializers.py
secult: This one may be reverted back

doc/changelog.rst
secult: Indeed

api/node/vm/utils.py
secult: It gives an explicit information to developer about the object expected in the variable, and this object is a list.

PR #249, opened 2017-09-06
Fixed bug when deploying VM with dhcp_passthrough network

dn0:


api/vm/base/utils.py
```diff
@@ -39,6 +39,9 @@ def _reset_allowed_ip_usage(vm, ip):
 def _is_ip_ok(ip_queryset, vm_ip, vm_network_uuid):
     """Helper function used below. Return True if vm_ip (string) is found in the IPAddress queryset and has the
```
secult: Consider removing this docstring

PR #254, opened 2017-09-07
Disks are defined and therefore we don't create but update disk#247

### LINKED ISSUE #247: Size parameter in template has no effect on new SunOS zone
```
es create /template/test-zone -vm_define 'json::{"vcpus":1,"ram":128,"ostype":5}' -vm_define_disk 'json::[{"size":2048,"image":"base-64-es"}]' 

es create /vm/test/define -template test-zone
```
The size will be the default value of the image and not the one defined in the template.

secult:
Closes #247

### DIFF
--- a/api/vm/define/vm_define.py
+++ b/api/vm/define/vm_define.py
@@ -96,17 +96,16 @@ class VmDefineView(VmDefineBaseView):
                                     'Error: %s', res.status_code, i, vm, vm.template, res.data)
                         break
 
-        if not vm.json_get_nics():
-            request = set_request_method(self.request, 'POST')
-            vm_define_nic = VmDefineNicView(request)
-            for i, data in enumerate(vm.template.vm_define_nic):
-                if data:
-                    logger.info('Creating nic_id=%d for vm %s defined by template %s', i, vm, vm.template)
-                    res = vm_define_nic.post(vm, i, data)
-                    if res.status_code != scode.HTTP_201_CREATED:
-                        logger.warn('Failed (%s) to add nic_id=%s into vm %s defined by template %s. '
-                                    'Error: %s', res.status_code, i, vm, vm.template, res.data)
-                        break
+        request = set_request_method(self.request, 'POST')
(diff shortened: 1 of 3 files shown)


secult:
I have a suspicion that there is the same problem with the nic parameters if they get defined both in the template and through vm_define.

api/vm/define/vm_define.py
```diff
@@ -75,26 +75,25 @@ def get_diff(self, vm, full=False):
...
-                                        'Error: %s', res.status_code, i, vm, vm.template, res.data)
-                    else:
-                        logger.info('Creating disk_id=%d for vm %s defined by template %s', i, vm, vm.template)
-                        res = vm_define_disk.post(vm, i, data)
-                        if res.status_code != scode.HTTP_201_CREATED:
-                            logger.warn('Failed (%s) to add disk_id=%s into vm %s defined by template %s. '
-                                        'Error: %s', res.status_code, i, vm, vm.template, res.data)
-                            break
+        # WARNING: This will temporary change the request.method to PUT as we are updating existing vm?
+        request = set_request_method(self.request, 'PUT')
```
dn0: I don't think that this will work. The `PUT` method should be used only for non-KVM VMs and only for the 1st disk; Otherwise it should be `POST`.

api/vm/define/vm_define.py
```diff
@@ -167,12 +168,8 @@ def put(self, vm, data, task_id=None, **kwargs):
                 # Task event for GUI
                 VmDefineHostnameChanged(self.request, vm, ser.old_hostname).send()
 
-            try:
-                return SuccessTaskResponse(self.request, ser.data, vm=vm, task_id=task_id,
-                                           msg=LOG_DEF_UPDATE, detail_dict=ser.detail_dict())
-            finally:
-                if ser.template_changed:
```
dn0: This is great, however we should probably raise an `ValidationError` in the serializer when the template change is detected.

PR #257, opened 2017-09-11
Added config to customize rabbitmq-server systemd service

mbag:
rabbitmq-server will be restarted automatically if it fails
This is related to issue erigones/esdc-factory#71

### DIFF
--- a/ans/upgrade/mgmt/v2.6.4/main.yml
+++ b/ans/upgrade/mgmt/v2.6.4/main.yml
@@ -13,3 +13,20 @@
     DJANGO_SETTINGS_MODULE: "core.settings"
   register: internal_images_update_result
   changed_when: "'Updated image' in internal_images_update_result.stdout"
+
+- name: Create directory for custom rabbit-systemd settings
+  file: path=/etc/systemd/system/rabbitmq-server.service.d
+        owner=root
+        group=root
+        mode=0755
+        state=directory
+
+- name: Copy cusotm configuration file for rabbitmq-service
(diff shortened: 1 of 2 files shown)


ans/upgrade/mgmt/v2.6.4/main.yml
```diff
@@ -13,3 +13,20 @@
...
+
+- name: Create directory for custom rabbit-systemd settings
+  file: path=/etc/systemd/system/rabbitmq-server.service.d
+        owner=root
+        group=root
+        mode=0755
+        state=directory
+
+- name: Copy cusotm configuration file for rabbitmq-service
+  copy: src= {{ current_task_dir }}/files/custom.conf
```
YanChii: - Ansible strongly asks for quotes when adding variables into string.
- Please remove the whitespace after `src=`.

ans/upgrade/mgmt/v2.6.4/main.yml
```diff
@@ -13,3 +13,20 @@
...
+        dest=dest=/etc/systemd/system/rabbitmq-server.service.d/custom.conf
+        owner=root
+        group=root
+        mode=0644
+
+- name: Reload systemd configuration files
+  shell: systemctl daemon-reload
```
YanChii: Please move this reload [here](https://github.com/erigones/esdc-ce/blob/master/ans/upgrade/lib/runupgrade.yml#L24) and add `notify` action to the `copy` task.

PR #262, opened 2017-09-20
261 post registration email

ricco386:
closes issue #261

### DIFF
--- a/api/email.py
+++ b/api/email.py
@@ -30,7 +30,7 @@ class SMTPEmailBackend(EmailBackend):
 
 
 def send_mail(subject, body, recipient_list, bcc_list=None, from_email=None, connection=None, attachments=None,
-              fail_silently=False, headers=None, cc_list=None, dc1_settings=None):
+              fail_silently=False, headers=None, cc_list=None, dc1_settings=None, content_subtype=None):
     """
     Like https://docs.djangoproject.com/en/dev/topics/email/#send-mail
     Attachment is a list of tuples (filename, content, mime_type), where mime_type can be None.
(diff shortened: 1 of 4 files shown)


api/email.py
```diff
@@ -64,6 +64,9 @@ def send_mail(subject, body, recipient_list, bcc_list=None, from_email=None, con
     else:
         logger.info('Sending mail to "%s" with subject "%s"', recipient_list, subject)
 
+    if content_subtype:
```
dn0: Please move this code block before the logger statement. The `Sending mail...` message should be displayed as late as possible.

api/email.py
```diff
@@ -129,6 +132,10 @@ def _sendmail(user, subject_template_name, body_template_name, recipient_list=No
     if extra_context is not None:
         context.update(extra_context)
 
+    content_subtype = None
```
dn0: Please move this to an `else` block.

api/email.py
```diff
@@ -129,6 +132,10 @@ def _sendmail(user, subject_template_name, body_template_name, recipient_list=No
...
+    if body_template_name.endswith('html'):
```
dn0: `.html`

gui/accounts/views.py
```diff
@@ -105,6 +108,23 @@ def registration_check(request, uidb64=None, token=None):
         }
         sms_sent = send_sms(profile.phone, msg)
 
+        template_path = dc_settings.PROJECT_DIR + '/gui/templates/'
```
dn0: Please use settings.PROJECT_DIR instead of `dc_settings`.
`path.join(settings.PROJECT_DIR, 'gui', 'templates')`

PR #267, opened 2017-09-21
Use newly collected nictags in the code

### LINKED ISSUE #227: Remove VMS_NET_NIC_TAGS -> switch to a real list of nic tags
The list will based on NIC tags collected after #226 is implemented.
It can be cached in redis.

mbag:
* removed VMS_NET_NIC_TAGS from code
* added classmethod to Node all_nictags that returns dict of all nictags and nictag types {name:type}
* added classmethod to Node all_nictags_choices that generates choices field for NetworkSerializer
* added property to Subnet nic_tag_type returning type of the nictag for Subnet

closes #227


vms/models/node.py
```diff
@@ -599,8 +629,9 @@ def save(self, sync_json=False, update_resources=True, zpool_update=False, zpool
             if save_ip and self._ip:
                 self._ip.save()
 
-        if clear_cache or status_changed:
-            self.all(clear_cache=True)
+            if clear_cache or status_changed:
```
dn0: Please add a comment explaining that this may eventually block the creation or update of the node from sysinfo.

vms/models/subnet.py
```diff
@@ -96,6 +97,12 @@ def set_resolvers(self, value):
     resolvers_api = property(get_resolvers, set_resolvers)
 
     @property
+    def nic_tag_type(self):
```
dn0: This should be added into the serializer used in net_manage.

mbag: As part of this PR, or in the next task regarding overlays? Because net_manage will be the next step.

dn0: net_manage is affected by this PR (removing NIC_TAGS... setting).

vms/models/node.py
```diff
@@ -406,6 +408,34 @@ def all(cls, clear_cache=False):
...
+
+                    nictags[nic['name']] = nic['type']
+
+            cache.set(NODES_ALL_KEY, nictags, NICTAGS_ALL_EXPIRES)
+
+        return nictags
+
+    @classmethod
+    def all_nictags_choices(cls):
+        """Return set of tuples that are used as choices in nictag field in NetworkSerializer"""
```
dn0: This returns a list and the set comprehension is probably not necessary since we don't allow duplicates anyway...

mbag:
@dn0 Here is the documentation update related to this change https://github.com/erigones/esdc-docs/pull/26

PR #273, opened 2017-09-26
Dc parameter conditionally required #265

### LINKED ISSUE #265: Missing "dc parameter required" validation when dc_bound parameter is true
When _dc_bound_ parameter is set to false, everything works correctly. However, when _dc_bound_ parameter is true, API does not strictly require _dc_ parameter. As a consequence, omitting the _dc_ parameter results in undefined behavior.
Affected calls:
POST net_manage, image_manage, template_manage, iso_manage, dns_domain_manage, maybe more

secult:
Closes #265

### DIFF
--- a/api/serializers.py
+++ b/api/serializers.py
@@ -1446,11 +1446,13 @@ class ConditionalDCBoundSerializer(InstanceSerializer):
 
     def _validate_dc_bound(self, value):
         if value:
-            if hasattr(self.object, 'get_dcs'):
-                dcs = self.object.get_dc_count()
-            else:
+            if hasattr(self.object, 'get_related_datacenters'):
+                dcs = self.object.get_related_datacenters()
+            elif hasattr(self.object, 'dc'):
                 dcs = self.object.dc.all()
-
+            else:
+                raise AssertionError(
+                    '%s has to implement either get_dcs method or have a dc relation.' % self._model_verbose_name)
             dcs_len = len(dcs)
 
             if dcs_len == 1:
--- a/api/serializers.py
+++ b/api/serializers.py
@@ -1445,15 +1445,9 @@ class ConditionalDCBoundSerializer(InstanceSerializer):
     dc_bound = BooleanField(source='dc_bound_bool', default=True)
 
     def _validate_dc_bound(self, value):
-        from gui.models import Role
-        from pdns.models import Domain
-        from vms.models import Dc
-
         if value:
-            if isinstance(self.object, Domain):
-                dcs = Dc.objects.filter(domaindc__domain_id=self.object.id)
-            elif isinstance(self.object, Role):
-                dcs = Dc.objects.filter(roles=self.object)
+            if hasattr(self.object, 'get_dcs'):
+                dcs = self.object.get_dc_count()
             else:
                 dcs = self.object.dc.all()
 
--- a/api/serializers.py
+++ b/api/serializers.py
@@ -1446,8 +1446,8 @@ class ConditionalDCBoundSerializer(InstanceSerializer):
 
     def _validate_dc_bound(self, value):
         if value:
-            if hasattr(self.object, 'get_related_datacenters'):
-                dcs = self.object.get_related_datacenters()
+            if hasattr(self.object, 'get_related_dcs'):
+                dcs = self.object.get_related_dcs()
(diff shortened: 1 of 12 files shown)


doc/changelog.rst
```diff
@@ -12,6 +12,7 @@ Features
 Bugs
 ----
 
+- Restricted dc_bound API calls to require datacenter to be explicitly set. - `#265 <https://github.com/erigones/esdc-ce/issues/265>`__
```
dn0: * Would be nice to mention the ``dc` parameter
* Remove dot at end of the sentence

api/serializers.py
```diff
@@ -1432,3 +1435,59 @@ def save(self, **kwargs):
...
 
         return self.object
+
+    @property
+    def _model_verbose_name(self):
+        # noinspection PyProtectedMember 2x
+        return self._model_._meta.verbose_name
+
+
+class ConditionalDCBoundSerializer(InstanceSerializer):
```
dn0: A docstring with explanation where/how this class is used would be helpful.

api/serializers.py
```diff
@@ -65,6 +65,9 @@
 from api.relations import *
 from api.fields import *
 from api.fields import is_simple_callable, get_component
+from gui.models import Role
```
dn0: This seems wrong. There should not be any imports from models here.

Please move it into the function/method.

api/serializers.py
```diff
@@ -1432,3 +1435,59 @@ def save(self, **kwargs):
...
+    def _model_verbose_name(self):
+        # noinspection PyProtectedMember 2x
+        return self._model_._meta.verbose_name
+
+
+class ConditionalDCBoundSerializer(InstanceSerializer):
+    _dc_bound = None
+    dc_bound = BooleanField(source='dc_bound_bool', default=True)
+
+    def validate(self, attrs):
```
dn0: Please move the `validate()` to be the last one (after field validators).

api/serializers.py
```diff
@@ -1432,3 +1435,59 @@ def save(self, **kwargs):
...
+        if attrs.get('dc_bound_bool', self.object.dc_bound_bool) and not self.request.data.get('dc', None):
+            err = {'model': self._model_verbose_name.lower()}
+            self._errors['dc_bound'] = self._errors['dc'] = (
+                _('You have to specify to which datacenter shall the %(model)s be bound. '
+                  'Either use the *dc* parameter or set the *dc_bound* parameter to False.') % err
```
dn0: We don't usually put hints into error messages, but I like it. However, since this is a JSON API, we should use `false` instead of `False` and I think that the parameters should be quoted in double quotes instead of stars (`*` -> `"`).

api/serializers.py
```diff
@@ -1432,3 +1435,59 @@ def save(self, **kwargs):
...
+        # noinspection PyProtectedMember 2x
+        return self._model_._meta.verbose_name
+
+
+class ConditionalDCBoundSerializer(InstanceSerializer):
+    _dc_bound = None
+    dc_bound = BooleanField(source='dc_bound_bool', default=True)
+
+    def validate(self, attrs):
+        if attrs.get('dc_bound_bool', self.object.dc_bound_bool) and not self.request.data.get('dc', None):
```
dn0: Should this check also affect other requests than POST (object creation)?

Please, lets think about this thoroughly.

api/serializers.py
```diff
@@ -1432,3 +1435,59 @@ def save(self, **kwargs):
...
+            err = {'model': self._model_verbose_name.lower()}
+            self._errors['dc_bound'] = self._errors['dc'] = (
```
dn0: Please use `s.ErrorList()`.

api/serializers.py
secult: The code using it seems wrong

gui/models/role.py
```diff
@@ -86,3 +86,7 @@ def web_data(self):
             'permissions': self.permissions_api,  # Not using values_list, because permissions.all() is prefetched
             'dcs': self.dcs_api,
         }
+
+    def get_related_datacenters(self):
```
dn0: This is a great idea 😄 . I would call it `get_related_dcs()` because currently all methods which return Dc objects have the `dcs` string (we don't use the `datacenter` name in code).

PR #281, opened 2017-10-23
Added overlay/VXLAN support to net_manage

### LINKED ISSUE #228: Add overlay/VXLAN support to net_manage
After #227 is implemented, we should add a vxlan DB field into the `Subnet` model and to the net_manage API call. In the API call (and GUI), the`vxlan` must be set to a number  when a *overlay* `nic_tag` is selected, otherwise it should be NULL (`None`).

The overlay NIC tag + VXLAN ID will be used during `vm_nic_define` configuration.

Add MTU parameter to `Subnet` model and `net_manage` - this will be inherited by VM NIC during `vm_define_nic`.

Followed by #286

mbag:
Closes #228


api/network/base/serializers.py
```diff
@@ -135,6 +155,10 @@ def validate(self, attrs):
             if dc_settings.VMS_NET_VLAN_RESTRICT and vlan_id not in dc_settings.VMS_NET_VLAN_ALLOWED:
                 self._errors['vlan_id'] = s.ErrorList([_('VLAN ID is not available in datacenter.')])
 
+            if dc_settings.VMS_NET_VXLAN_RESTRICT and vxlan_id not in dc_settings.VMS_NET_VXLAN_ALLOWED:
+                self._errors['vxlan_id'] = s.ErrorList([_('VXLAN ID is not available in datacenter.')])
+
+
```
dn0: PEP8: too many blank lines.

api/network/base/serializers.py
```diff
@@ -122,7 +133,16 @@ def validate(self, attrs):
 
             if limit is not None:
                 if Subnet.objects.filter(dc_bound=self._dc_bound).count() >= int(limit):
-                    raise s.ValidationError(_('Maximum number of networks reached'))
+                    raise s.ValidationError(_('Maximum number of networks reached.'))
+
+        # retrieve all available nictags and see what is the type of the current nic tag
+        # if type is overlay then vxlan is mandatory argument
+        if Node.all_nictags()[nic_tag] == 'overlay':
```
dn0: `overlay` or `overlay rule`?

api/network/base/serializers.py
```diff
@@ -122,7 +140,25 @@ def validate(self, attrs):
...
+        # if type is overlay then vxlan is mandatory argument
+        if nic_tag_type == 'overlay rule':
+            if not vxlan_id:
+                self._errors['vxlan_id'] = s.ErrorList([_('VXLAN ID is required when an '
+                                                          'overlay NIC tag is selected.')])
+        else:
+            attrs['vxlan_id'] = None
+
+        # validate MTU for overlays and etherstubs, and physical nics
+        if nic_tag_type in ('overlay rule') and not mtu:
```
dn0: Invalid if condition.

gui/templates/gui/dc/network_list.html
```diff
@@ -135,12 +135,14 @@
...
               <a href="{% url 'dc_network_ip_list' net.name %}?{{ qs }}" class="btn-link" title="{% trans "Show/edit network IP addresses" %}" data-toggle="tooltip" data-placement="bottom">{{ net.network|cidr:net.netmask }}</a>
+              {% if net.gateway %}<br><small>{% trans "via" %}&nbsp;{{ net.gateway }}</small>{% endif %}
               {% if net.dhcp_passthrough %}&nbsp;
                 <span title="{% trans "IP addresses for this network are managed by an external service" %}" data-toggle="tooltip" data-placement="bottom"><i class="icon-info-sign"></i></span>
               {% endif %}
             </td>
-            <td class="hidden-phone nowrap">{{ net.gateway }}</td>
             <td class="nowrap">{{ net.vlan_id }}</td>
+            <td class="hidden-phone nowrap">{{ net.nic_tag }}{% if net.vxlan_id %}/{{ net.vxlan_id }}{% endif %}</td>
+            <td class="hidden-phone nowrap">{{ net.nic_tag }}{% if net.mtu %}/{{ net.mtu }}{% endif %}</td>
```
dn0: There won't be a MTU column in network list.

gui/templates/gui/dc/network_admin_form.html
```diff
@@ -14,6 +14,8 @@
 {% include "gui/form_field.html" with field=form.gateway %}
 {% include "gui/form_field.html" with field=form.vlan_id %}
 {% include "gui/form_field.html" with field=form.nic_tag %}
+{% include "gui/form_field.html" with field=form.vxlan_id %}
+{% include "gui/form_field.html" with field=form.mtu %}
```
dn0: MTU field should be in the advanced section.

api/network/base/serializers.py
```diff
@@ -122,7 +140,25 @@ def validate(self, attrs):
...
+        else:
+            attrs['vxlan_id'] = None
+
+        # validate MTU for overlays and etherstubs, and physical nics
+        if nic_tag_type == 'overlay rule' and not mtu:
+            # if MTU was not set for the overlay
+            attrs['mtu'] = 1400
+
+        if nic_tag_type in ('normal', 'aggr') and mtu and mtu < 1500:
+            self._errors['mtu'] = s.ErrorList([_('MTU must be from integer interval [1500, 9000].')])
```
dn0: I would say: ...must be larger than 1500.

mbag: "must be larger than 1500" does not mention anything about upper limit. The above is fairly standard mathematical notation, on how to tell someone that an integer should be from certain interval. If square brackets are the problem due to the fact that they mean something different in python, we can use for example "between 1500 and 9000"

dn0: What about?
```python
s.ErrorList([s.IntegerField.default_error_messages['min_value'] % {'limit_value': 1500}])
```

mbag: That could work. However, I see that in the future someone will complain that sometimes 576 can be entered as lower limit and sometimes 1500. The fact that the type of the netwrok is not mentioned in error message might confuse users.

dn0: Agreed. Either is should be mentioned in the documentation (API docs, user guide) or in the error message.

vms/models/vm.py
```diff
@@ -201,7 +201,7 @@ class Vm(_StatusModel, _JsonPickleModel, _OSType, _UserTasksModel):
     RESERVED_MDATA_KEYS = frozenset(['resize_needed'])  # Bug #chili-721
 
     _DISKS_REMOVE_EMPTY = ()
-    _NICS_REMOVE_EMPTY = (('gateway', ''), ('allowed_ips', ()))
+    _NICS_REMOVE_EMPTY = (('gateway', ''), ('allowed_ips', ()), ('mtu', None))
```
dn0: This break NIC updates. Use case:

1. add a vNIC (1) with a network that has an MTU set.
2. deploy VM
3. update the same vNIC (1) to have another network (without MTU set)

We need to find a way to remove mtu via `vmadm update`. Maybe setting to `''` empty string will do this.

dn0: Disabled NIC update for these kind of situations -> user has to delete/add NIC.

PR #282, opened 2017-10-23
Added VMS_VM_DEFINE_LIMIT

### LINKED ISSUE #280: Add limit for number of VMs in a virtual datacenter
This will be a DC setting. It will be checked when a new VM definitions is created.

Don't forget to update esdc-docs.

dn0:
Closes #280


api/dc/base/serializers.py
```diff
@@ -271,6 +272,10 @@ class DcSettingsSerializer(s.InstanceSerializer):
     VMS_ZONE_ENABLED = s.BooleanField(label='VMS_ZONE_ENABLED',  # Module
                                       help_text=_('Whether to enable support for SunOS and Linux zones in '
                                                   'this virtual datacenter.'))
+
+    VMS_VM_DEFINE_LIMIT = s.IntegerField(label='VMS_VM_DEFINE_LIMIT', required=False,
```
mbag: Should negative limit be allowed? If not use PositiveIntegerField.

mbag: Also, when setting limits, it should be validated against current number of servers in the virtual data center. We should warn admin if current number of defined servers in data center is higher, than the newly value that admin is trying to set.

dn0: Currently no `*_LIMIT` setting has a check for negative numbers. I agree that negative values don't make any sense. But I prefer to keep it consistent here. It certainly won't break things because the limit is only used in comparisons.

I can imagine a situation where a user wants to have a certain amount of VMs in a DC, but currently has more VMs deployed. Now he can do that and wait for the DCAdmin to remove other VMs.

PR #285, opened 2017-10-31
Enable access to snapshots from SunOS/LX zone 

### LINKED ISSUE #284: Enable access to snapshots from SunOS zone
(Moved from https://github.com/erigones/smartos-live/issues/2)

PR #285

dn0:
Related to #284


YanChii:
Will the snapshots be mounted also after the zone stop+start?
Otherwise ok for me.
J.

dn0:
@YanChii mounting/umounting during start/stop works thanks to this code: https://github.com/joyent/smartos-live/blob/master/overlay/generic/usr/lib/brand/jcommon/statechange#L587
However, it looks like that it won't work for delegated dataset.

PR #288, opened 2017-11-08
Updated the VM templates API + removed *experimental* flag from the `template_manage` API call

### LINKED ISSUE #256: Improve template creation so that the feature may not be experimental anymore
Improvement suggestions:
- check for ostype mismatch among ostype, vm_define ostype and disk image os type
- remove "experimental feature" documenation flag
- remove the template deletion restriction ("Template is used by some VMs" in 2 places).

dn0:
Closes #256

- Checking for possible mismatch between ``template.ostype`` and ``template.vm_define.ostype``.
- Added note about removing used templates, networks and disk images.
- Added example for ``POST /template/(name)`` into the API documentation.

### DIFF
--- a/api/template/base/serializers.py
+++ b/api/template/base/serializers.py
@@ -145,7 +145,22 @@ class TemplateSerializer(s.ConditionalDCBoundSerializer):
 
             if limit is not None:
                 if VmTemplate.objects.filter(dc_bound=self._dc_bound).count() >= int(limit):
-                    raise s.ValidationError(_('Maximum number of server templates reached'))
+                    raise s.ValidationError(_('Maximum number of server templates reached.'))
+
+        try:
+            ostype = attrs['ostype']
+        except KeyError:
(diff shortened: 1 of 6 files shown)


api/template/base/serializers.py
```diff
@@ -145,7 +145,22 @@ def validate(self, attrs):
 
             if limit is not None:
                 if VmTemplate.objects.filter(dc_bound=self._dc_bound).count() >= int(limit):
-                    raise s.ValidationError(_('Maximum number of server templates reached'))
+                    raise s.ValidationError(_('Maximum number of server templates reached.'))
+
+        try:
```
secult: Please add a short rationale about this code - why we cannot automatically take the template ostype or the other way around

dn0: Please check the comment below.

dn0:
@secult If you're happy with the PR, you can merge it and close #256.

PR #290, opened 2017-11-09
289 platform upgrade

### LINKED ISSUE #289: Add platform upgrade command
`esdc-platform-upgrade-diskboot`
```
Usage:   ./esdc-platform-upgrade-diskboot <new_platform_version> [--keep-smf-db]
Example: ./esdc-platform-upgrade-diskboot 20170624T192838Z
Args:
  --keep-smf-db         don't clear SMF database (useful if you don't want to loose
                        your manual changes to any SMF service properties)
```
`esdc-platform-upgrade-usbboot`
```
Usage:   ./esdc-platform-upgrade-usbboot <new_esdc_version>
Example: ./esdc-platform-upgrade-usbboot v2.6.7
```
The whole upgrade procedure can be managed like this:
```
cat /opt/erigones/bin/esdc-platform-upgrade
#!/bin/bash

if beadm list -H | grep -q dcos-; then
    /opt/erigones/bin/esdc-platform-upgrade-diskboot $PLATFORM_VER
elif /bin/bootparams | grep -q "^netboot=true"; then
    echo "This node is network booted. Nothing to do."
    exit 1
else
    /opt/erigones/bin/esdc-platform-upgrade-usbboot $ESDC_VER
fi
```

YanChii:
Closes  #289


bin/eslib/esdc-platform-upgrade-disk
```diff
@@ -0,0 +1,154 @@
...
+			;;
+		*)
+			echo "Ignoring unknown argument '${1}'"
+			;;
+	esac
+	shift
+done
+
+if [[ "x${FORCE}" != "x1" ]] && [[ "${PLATFORM_VER}" == "$(uname -v | sed 's/^[a-z]*_//')" ]]; then
+	echo "The requested platform version is already running. Aborting upgrade."
```
dn0: Consider using `die()` from functions.sh.

bin/eslib/esdc-platform-upgrade-disk
```diff
@@ -0,0 +1,154 @@
...
+	lofi_remove "${USR_LOOPDEV}"
+	umount_path "${PLATFORM_MOUNT_DIR}"
+	lofi_remove "${PLATFORM_LOOPDEV}"
+	rm -rf "${UPG_DIR}"
+}
+upgrade_failed() {
+	printmsg "Upgrade failed!"
+	post_cleanup
+}
+trap upgrade_failed EXIT
```
dn0: Don't put a `trap` between function definitions. Please maintain a common order in a script (0. set 1. declarations/constants, 2. function definitions (internal first), ...)

The `trap` should go after the `### START ###`.

bin/eslib/esdc-platform-upgrade-disk
```diff
@@ -0,0 +1,154 @@
...
+
+printmsg() {
+	echo "*** $* ***"
+}
+
+### START ###
+
+# pre-cleanup
+rm -rf "${UPG_DIR}"
```
dn0: This seems a bit nasty. Maybe it would be better to just fail if `/opt/upgrade` exists. If it exists it is not from this script (trap) and we should not delete user data without asking.

bin/eslib/esdc-platform-upgrade-disk
```diff
@@ -0,0 +1,154 @@
...
+
+# correct the dcos-1
+# (issue https://github.com/erigones/esdc-factory/issues/94)
+if [[ "$(zfs get -Ho value canmount zones/ROOT/dcos-1)" != "noauto" ]]; then
+	zfs set canmount=noauto zones/ROOT/dcos-1
+fi
+
+# start download
+printmsg "Downloading the new platform"
+curl -k ${CURL_OPTS} "${PLATTFORM_DOWNLOAD_URL}" > "${PLATFORM_FILE}"
```
dn0: Use `-m` with some reasonable value (15?). Or `--connect-timeout`.
Use `-o` instead of `>`.
I also suggest adding the `-L` switch (support for future redirects).
Certificates on SmartOS work, so `-k` is not needed, right?

bin/eslib/esdc-platform-upgrade-disk
```diff
@@ -0,0 +1,154 @@
...
+if [[ "${KEEP_SMF_DB}" != "1" ]]; then
+	# clear SMF database so it can be recreated at reboot
+	rm -f "${DCOS_MNTDIR}/etc/svc/repository.db"
+fi
+
+printmsg "Activating the new boot environment at next reboot"
+beadm activate "${NEW_BE}"
+trap post_cleanup EXIT
+printmsg "Upgrade completed successfuly"
+printmsg "Please reboot as soon as possible.."
```
dn0: Two dots at end of sentence.

bin/eslib/esdc-platform-upgrade-disk
```diff
@@ -0,0 +1,154 @@
...
+cksum -x sha1 "${DCOS_MNTDIR}/platform/i86pc/amd64/boot_archive" | cut -d' ' -f1 > "${DCOS_MNTDIR}/platform/i86pc/amd64/boot_archive.hash"
+chown -R root:root "${DCOS_MNTDIR}/platform/i86pc/amd64"
+if [[ "${KEEP_SMF_DB}" != "1" ]]; then
+	# clear SMF database so it can be recreated at reboot
+	rm -f "${DCOS_MNTDIR}/etc/svc/repository.db"
+fi
+
+printmsg "Activating the new boot environment at next reboot"
+beadm activate "${NEW_BE}"
+trap post_cleanup EXIT
```
dn0: There must a better way.  Changing the trap complicates readability.
Please add only one trap and show the "Upgrade failed" message only if the upgrade fails.

bin/eslib/esdc-platform-upgrade-disk
```diff
@@ -0,0 +1,154 @@
...
+
+printmsg "Creating a new boot environment"
+BEADM_NEW_NUMBER="$(expr 1 + $(beadm list -H | tail -1 | cut -d';' -f1 | cut -d- -f2))"
+if ! [[ $BEADM_NEW_NUMBER =~ ^[0-9]+$ ]]; then
+	# BEADM_NEW_NUMBER is not a correct number
+	echo "Cannot get number of current BE."
+	exit 5
+elif beadm list -H | cut -d';' -f1 | grep -q "dcos-${BEADM_NEW_NUMBER}"; then
+	# current BE already exists (beadm lists BE out of order?)
+	echo "Cannot get name of new BE."
```
dn0: Consider using `die()`

bin/eslib/esdc-platform-upgrade-disk
```diff
@@ -0,0 +1,154 @@
...
+PLATFORM_LOOPDEV="$(lofiadm -a "${BOOT_ARCHIVE}")"
+mount -F ufs "${PLATFORM_LOOPDEV}" "${PLATFORM_MOUNT_DIR}"
+USR_LOOPDEV="$(lofiadm -a "${USR_ARCHIVE}")"
+mount -r -F ufs "${USR_LOOPDEV}" "${USR_ARCHIVE_MOUNT_DIR}"
+
+printmsg "Creating a new boot environment"
+BEADM_NEW_NUMBER="$(expr 1 + $(beadm list -H | tail -1 | cut -d';' -f1 | cut -d- -f2))"
+if ! [[ $BEADM_NEW_NUMBER =~ ^[0-9]+$ ]]; then
+	# BEADM_NEW_NUMBER is not a correct number
+	echo "Cannot get number of current BE."
```
dn0: Consider using `die()`

bin/eslib/esdc-platform-upgrade-disk
```diff
@@ -0,0 +1,154 @@
...
+tar zxf "${PLATFORM_FILE}" -C "${PLATFORM_DIR}"
+
+printmsg "Accessing the new platform files"
+PLATFORM_LOOPDEV="$(lofiadm -a "${BOOT_ARCHIVE}")"
+mount -F ufs "${PLATFORM_LOOPDEV}" "${PLATFORM_MOUNT_DIR}"
+USR_LOOPDEV="$(lofiadm -a "${USR_ARCHIVE}")"
+mount -r -F ufs "${USR_LOOPDEV}" "${USR_ARCHIVE_MOUNT_DIR}"
+
+printmsg "Creating a new boot environment"
+BEADM_NEW_NUMBER="$(expr 1 + $(beadm list -H | tail -1 | cut -d';' -f1 | cut -d- -f2))"
```
dn0: This code block should be executed sooner. No need to download the new platform if you cannot calculate the BE name.

bin/eslib/esdc-platform-upgrade-disk
```diff
@@ -0,0 +1,154 @@
...
+beadm mount "${NEW_BE}" "${DCOS_MNTDIR}"
+
+printmsg "Updating the new boot environment in the background"
+rsync -ac --delete --exclude=boot/loader.conf "${PLATFORM_MOUNT_DIR}/"{bin,boot,kernel,lib,platform,sbin,smartdc} "${DCOS_MNTDIR}"
+rsync -ac --delete "${USR_ARCHIVE_MOUNT_DIR}/"  "${DCOS_MNTDIR}/usr/"
+rsync -a --delete "${PLATFORM_DIR}/platform-${PLATFORM_VER}/i86pc/amd64" "${DCOS_MNTDIR}/platform/i86pc"
+# update boot_archive checksum
+cksum -x sha1 "${DCOS_MNTDIR}/platform/i86pc/amd64/boot_archive" | cut -d' ' -f1 > "${DCOS_MNTDIR}/platform/i86pc/amd64/boot_archive.hash"
+chown -R root:root "${DCOS_MNTDIR}/platform/i86pc/amd64"
+if [[ "${KEEP_SMF_DB}" != "1" ]]; then
```
dn0: Use integer comparison here.
Add a newline before the `if` statement to improve readability.

bin/eslib/esdc-platform-upgrade-disk
```diff
@@ -0,0 +1,154 @@
...
+
+NEW_BE="dcos-${BEADM_NEW_NUMBER}"
+beadm create "${NEW_BE}"
+beadm mount "${NEW_BE}" "${DCOS_MNTDIR}"
+
+printmsg "Updating the new boot environment in the background"
+rsync -ac --delete --exclude=boot/loader.conf "${PLATFORM_MOUNT_DIR}/"{bin,boot,kernel,lib,platform,sbin,smartdc} "${DCOS_MNTDIR}"
+rsync -ac --delete "${USR_ARCHIVE_MOUNT_DIR}/"  "${DCOS_MNTDIR}/usr/"
+rsync -a --delete "${PLATFORM_DIR}/platform-${PLATFORM_VER}/i86pc/amd64" "${DCOS_MNTDIR}/platform/i86pc"
+# update boot_archive checksum
```
dn0: I think that this comment can be also printed to the user.

bin/eslib/esdc-platform-upgrade-disk
```diff
@@ -0,0 +1,154 @@
...
+cksum -x sha1 "${DCOS_MNTDIR}/platform/i86pc/amd64/boot_archive" | cut -d' ' -f1 > "${DCOS_MNTDIR}/platform/i86pc/amd64/boot_archive.hash"
+chown -R root:root "${DCOS_MNTDIR}/platform/i86pc/amd64"
+if [[ "${KEEP_SMF_DB}" != "1" ]]; then
+	# clear SMF database so it can be recreated at reboot
```
dn0: This operation should be printed to the user.

bin/esdc-platform-upgrade
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -e
+
+ESLIB="$(dirname $0)/eslib"
+
+# choose the right upgrade path for this node (disk, net, usb)
+if beadm list -H | grep -q dcos-; then
```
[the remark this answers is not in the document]
YanChii: Refactored matching and creating new BE to be independent of the `dcos` prefix. Now the new BE name will be derrived from the current BE name regardless the prefix and number. 

If the trailing number exists, it will be incremented. 

If the trailing number does not exist, the `-1` will be added.

I've also added checks for BE name existence (in case we're upgrading from non-newest BE).

bin/esdc-platform-upgrade
```diff
@@ -0,0 +1,14 @@
...
+    "${ESLIB}/esdc-platform-upgrade-disk" "$@"
+elif /bin/bootparams | grep -q "^netboot=true"; then
+    echo "This node is network booted. Nothing to do."
+    exit 0
+else
```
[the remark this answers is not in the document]
YanChii: Additional checks added.

bin/eslib/esdc-platform-upgrade-disk
YanChii: Now it fails in `esdc-platform-upgrade` if it cannot get the BE name. If the BE name can be obtained, the new BE name will be successful for sure.

bin/eslib/esdc-platform-upgrade-disk
YanChii: CAs in Erigonos are not rolled out so AFAIK `-k` is needed.

YanChii:
Multiple functions refactored into `functions.sh` and made globally reusable.

bin/eslib/esdc-platform-upgrade-disk
YanChii: Introduced `FINISHED_SUCCESSFULLY` but personally I don't like it.

bin/eslib/esdc-platform-upgrade-disk
dn0: Just use `/opt/local/bin/curl` instead of `/usr/bin/curl -k`.

bin/esdc-platform-upgrade
```diff
@@ -0,0 +1,21 @@
...
+if dc_booted_from_hdd && [[ -n "$(_beadm_get_current_be_name)" ]]; then
+	# a boot env exists, this is disk install
+    "${ESLIB}/esdc-platform-upgrade-disk.sh" "$@"
+elif /bin/bootparams | grep -q "^netboot=true"; then
+    echo "This node is network booted. Nothing to do."
+    exit 0
+elif dc_booted_from_usb; then
+    "${ESLIB}/esdc-platform-upgrade-usb.sh" "$@"
+else
+	die 2 "Cannot determine platform boot type. Please contact support."
```
dn0: Wrong indentation.
Please remove the message about contacting support.

bin/eslib/functions.sh
```diff
@@ -52,6 +53,13 @@ BC=${BC:-"/usr/bin/bc"}
 QGA=${QGA:-"${ERIGONES_HOME}/bin/qga-client"}
 QGA_SNAPSHOT=${QGA_SNAPSHOT:-"${ERIGONES_HOME}/bin/qga-snapshot"}
 NODE=${NODE:-"/usr/node/bin/node"}
+BEADM=${BEADM:-"/usr/sbin/beadm"}
+RSYNC=${RSYNC:-"/usr/bin/rsync"}
+CURL=${CURL:-"/usr/bin/curl"}
```
dn0: Are you sure you want use `/usr/bin/curl` by default?

bin/eslib/functions.sh
```diff
@@ -163,6 +175,25 @@ round() {
 	printf "%.${places}f" "${number}"
 }
 
+umount_path() {
```
dn0: There is a convention how all functions are written here. Please follow the convention:

Every positional argument is locally saved to a named variable.

bin/eslib/functions.sh
```diff
@@ -163,6 +175,25 @@ round() {
...
+	if ${MOUNT} | grep -q "${1}"; then
```
dn0: This won't work in some cases, e.g. you have nested mountpoints.

bin/esdc-platform-upgrade
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+set -e
+
+ESLIB="$(dirname $0)/eslib"
+
+. "${ESLIB}/functions.sh"
```
dn0: The first thing to do is to determine ERIGONES_HOME. Then source it from there.
Please look at other scripts first. Consistency matters.

bin/eslib/functions.sh
YanChii: Yes.. in that case it will fail. This should be handled by user before calling of the function. Recursive umount should not happen unless explicitly specified (if it's really needed).

bin/eslib/esdc-platform-upgrade-disk.sh
```diff
@@ -0,0 +1,198 @@
...
+	shift
+done
+
+# check platform version format
+if [[ ! "${PLATFORM_VER}" =~ ^[0-9]+T[0-9]+Z$ ]] && [[ ! "${PLATFORM_VER}" =~ ^v[0-9]+\.[0-9]+ ]]; then
+	die 2 "Unknown format of platform version: ${PLATFORM_VER}"
+fi
+
+# query platform version from esdc version if needed
+if [[ "${PLATFORM_VER}" =~ ^v[0-9]+\.[0-9]+ ]]; then
```
dn0: We should also support version in other format - basically everything that begins with a v[0-9] should be considered a version string.

bin/eslib/esdc-platform-upgrade-disk.sh
```diff
@@ -0,0 +1,198 @@
...
+${BEADM} create "${NEW_BE}"
+${BEADM} mount "${NEW_BE}" "${DCOS_MNTDIR}"
+
+printmsg "Updating the new boot environment in the background"
+${RSYNC} -ac --delete --exclude=boot/loader.conf "${PLATFORM_MOUNT_DIR}/"{bin,boot,kernel,lib,platform,sbin,smartdc} "${DCOS_MNTDIR}"
+${RSYNC} -ac --delete "${USR_ARCHIVE_MOUNT_DIR}/"  "${DCOS_MNTDIR}/usr/"
+${RSYNC} -a --delete "${PLATFORM_DIR}/platform-${PLATFORM_VER}/i86pc/amd64" "${DCOS_MNTDIR}/platform/i86pc"
+printmsg "Update boot_archive checksum"
+checksum "${DCOS_MNTDIR}/platform/i86pc/amd64/boot_archive" > "${DCOS_MNTDIR}/platform/i86pc/amd64/boot_archive.hash"
+chown -R root:root "${DCOS_MNTDIR}/platform/i86pc/amd64"
```
dn0: Why is this needed?

bin/eslib/esdc-platform-upgrade-usb.sh
```diff
@@ -0,0 +1,125 @@
...
+	echo "Usage:   $0 <new_dc_version> [-v] [-f]"
+	echo "Example: $0 v2.6.7"
+	echo "Args:"
+	echo "  -v                verbose download"
+	echo "  -f                force upgrade even when the requested version is already installed"
+
+	exit 1
+fi
+
+if ! [[ "${ESDC_VER}" =~ ^v[0-9]\.[0-9]\.[0-9]$ ]]; then
```
dn0: This check is too strict. We should also support other version strings.

bin/eslib/esdc-platform-upgrade-usb.sh
```diff
@@ -0,0 +1,125 @@
...
+CURR_USB_VER="$(cat "${USB_VERSION_FILE}" | sed -e 's/^esdc-.e-.n-//')"
+# shellcheck disable=SC2002
+WANTED_USB_IMG_VARIANT="$(cat "${USB_VERSION_FILE}" | sed -re 's/^(esdc-.e-.n-).*/\1/')"
+if [[ ${FORCE} -ne 1 ]] && [[ "${CURR_USB_VER}" == "${NEW_USB_VER}" ]]; then
+	die 0 "Requested ESDC version is already on the USB key. Nothing to do."
+fi
+
+umount_usb_key
+
+printmsg "Downloading new platform"
```
dn0: new USB image

bin/eslib/esdc-platform-upgrade-usb.sh
```diff
@@ -0,0 +1,125 @@
...
+ESDC_IMG="${WANTED_USB_IMG_VARIANT}${NEW_USB_VER}.img"
+ESDC_IMG_FULL="${UPG_DIR}/${ESDC_IMG}"
+ESDC_DOWNLOAD_URL="https://download.erigones.org/esdc/usb/stable/${ESDC_IMG}.gz"
+
+trap cleanup EXIT
+
+mkdir -p "${UPG_DIR}"
+# shellcheck disable=SC2086
+${CURL} ${CURL_OPTS} ${CURL_DEFAULT_OPTS} -o "${ESDC_IMG_FULL}.gz" "${ESDC_DOWNLOAD_URL}"
+printmsg "Unpacking new platform"
```
dn0: platform -> USB image

bin/eslib/functions.sh
```diff
@@ -821,3 +858,156 @@ _service_instance_delete() {
...
+
+_beadm_get_active_be_name() {
+	${BEADM} list -H 2>/dev/null | "${AWK}" -F';' '{if($3 == "R" || $3 == "NR") {print $1}}'
+}
+
+_beadm_check_be_exists() {
+	local BE="${1}"
+
+	${BEADM} list -H 2>/dev/null | grep -q -- "^${BE};"
+	return $?
```
dn0: This return statement is redundant. Either

bin/eslib/functions.sh
```diff
@@ -821,3 +858,156 @@ _service_instance_delete() {
...
+}
+
+_beadm_get_next_be_name() {
+	CURR_BE="$(_beadm_get_current_be_name)"
```
dn0: missing local keyword + local variables should be lowercase

bin/eslib/functions.sh
```diff
@@ -821,3 +858,156 @@ _service_instance_delete() {
...
+_beadm_get_current_be_name() {
+	${BEADM} list -H 2>/dev/null | "${AWK}" -F';' '{if($3 == "N" || $3 == "NR") {print $1}}'
+}
+
+_beadm_get_active_be_name() {
+	${BEADM} list -H 2>/dev/null | "${AWK}" -F';' '{if($3 == "R" || $3 == "NR") {print $1}}'
+}
+
+_beadm_check_be_exists() {
+	local BE="${1}"
```
dn0: local variables should be lowercase

bin/eslib/functions.sh
```diff
@@ -821,3 +858,156 @@ _service_instance_delete() {
...
+	echo "/mnt/$(svcprop -p 'joyentfs/usb_mountpoint' svc:/system/filesystem/smartdc:default)"
+}
+
+mount_usb_key() {
+	if [[ -n "$(_usbkey_get_mounted_path)" ]]; then
+		# already mounted
+		return 0
+	fi
+
+	local ALLDISKS=`/usr/bin/disklist -a`
```
dn0: local variables should be lowercase.

bin/eslib/functions.sh
```diff
@@ -821,3 +858,156 @@ _service_instance_delete() {
...
+	local USBMNT="$(_usbkey_get_mountpoint)"
+
+	mkdir -p "${USBMNT}"
+	for key in ${ALLDISKS}; do
+		if [[ "$(${FSTYP} /dev/dsk/${key}p1 2> /dev/null)" == 'pcfs' ]]; then
+			${MOUNT} -F pcfs -o foldcase,noatime /dev/dsk/${key}p1 ${USBMNT};
+			if [[ $? == "0" ]]; then
```
dn0: use if command instead of $? comparision

bin/eslib/functions.sh
```diff
@@ -821,3 +858,156 @@ _service_instance_delete() {
...
+		return 0
+	fi
+
+	local ALLDISKS=`/usr/bin/disklist -a`
+	local USBMNT="$(_usbkey_get_mountpoint)"
+
+	mkdir -p "${USBMNT}"
+	for key in ${ALLDISKS}; do
+		if [[ "$(${FSTYP} /dev/dsk/${key}p1 2> /dev/null)" == 'pcfs' ]]; then
+			${MOUNT} -F pcfs -o foldcase,noatime /dev/dsk/${key}p1 ${USBMNT};
```
dn0: trailing ;

bin/eslib/esdc-platform-upgrade-disk.sh
YanChii: Because files in the platform are not root-owned. The owner is uid 1000.
```
root@n1:/var/tmp# ll platform-20170624T192838Z/i86pc/amd64/
total 546570
drwxr-xr-x   2 1000     1000           6 Jun 24 23:18 ./
drwxr-xr-x   4 1000     1000           4 Jun 24 23:12 ../
-rw-r--r--   1 1000     1000        266M Jun 24 23:17 boot_archive
-rw-r--r--   1 1000     1000        1.3K Jun 24 23:13 boot_archive.gitstatus
-rw-r--r--   1 1000     1000          41 Jun 24 23:18 boot_archive.hash
-rw-r--r--   1 1000     1000        1.2M Jun 24 23:13 boot_archive.manifest
```

dn0: Then this is a bug in our PI build system.

PR #292, opened 2017-11-13
95 Zabbix Alert Listing

### LINKED ISSUE #95: Alert listing (Zabbix events) in API and GUI
Add views for listing actual and historical alerts per VM, Node and DC (similar to task log list).

ricco386:
Added ability to list alerts from Zabbix. We have and asynchronous API call to query Zabbix and get alerts that are active (API support filtering). Added new GUI view and navigation to display active alerts. (GUI filtering is not done), even GUI is asynchronous, but for better UX we made it in a way that it looks synchronous, eg. user waits for the task to finish and that alerts are displayed.

Closes #95


api/mon/backends/zabbix/base.py
```diff
@@ -1,5 +1,6 @@
 from logging import getLogger, INFO, WARNING, CRITICAL, ERROR
 from time import time
+from datetime import datetime, timedelta
```
dn0: Unused import

api/mon/backends/zabbix/base.py
```diff
@@ -675,9 +719,9 @@ def _diff_interfaces(cls, obj, host, interface):
                     hi = iface
 
                     if (iface['dns'] != interface['dns']
-                            or iface['ip'] != interface['ip']
-                            or str(iface['port']) != str(interface['port'])
-                            or str(iface['useip']) != str(interface['useip'])):
+                        or iface['ip'] != interface['ip']
+                        or str(iface['port']) != str(interface['port'])
+                        or str(iface['useip']) != str(interface['useip'])):
```
dn0: My PyCharm is giving me a PEP8 error on this.

api/mon/backends/zabbix/base.py
```diff
@@ -898,6 +942,227 @@ def get_hostgroup_list(self):
...
+            return res
+        except (KeyError, IndexError) as e:
+            logger.exception(e)
+            raise ZabbixError(e)
+
+        # If trigger is lost (broken expression) we skip it
+        return (trigger for trigger in res if trigger['hosts'])
+
+    # noinspection PyUnusedLocal
+    def _show_alerts(self, since=None, until=None, last=None, display_notes=True, display_items=True,
```
dn0: This method should be public, because it is used by the main Zabbix() class.

api/mon/backends/zabbix/monitor.py
```diff
@@ -461,6 +461,17 @@ def hostgroup_list(self, prefix=''):
         """[EXTERNAL] Return list of available hostgroups"""
         return list(self._get_filtered_hostgroups(prefix=prefix))
 
+    def _get_filtered_alerts(self, *args, **kwargs):
+        """This is a generator function
+        """
```
dn0: Redundant newline in the comment

api/mon/base/serializers.py
```diff
@@ -0,0 +1,26 @@
...
+class AlertSerializer(s.Serializer):
+    """
+    Used by NodeHistoryView to validate zpools value.
+    """
+    since = s.TimeStampField(required=False)
+    until = s.TimeStampField(required=False)
+    last = s.IntegerField(required=False)
+    display_notes = s.BooleanField(default=True)
+    display_items = s.BooleanField(default=True)
+    hosts_or_groups = s.ArrayField(default=[])
```
dn0: It is nice that the user can filter for a specific host or group. But this filter should work on an already pre-selected list of servers (or a hostgroup) for a specific DC.

gui/mon/views.py
```diff
@@ -14,3 +21,50 @@ def monitoring_server(request):
...
+
+@login_required
+@ajax_required
+@profile_required
+def get_alert_from_zabbix(request):
+    context = collect_view_data(request, 'mon_alert_list')
+    method = 'GET'
+    logger.info('Calling API view %s mon_alert_list(%s, data=%s) by user %s in DC %s',
+                method, request, None, request.user, request.dc)
+    res = call_api_view(request, method, mon_alert_list, disable_throttling=True)
```
dn0: Why disable_throttling?

gui/templates/base.html
```diff
@@ -105,6 +105,7 @@
   <script src="{{ STATIC_URL }}gui/js/vm/vm-backup.js" type="text/javascript"></script>
   <script src="{{ STATIC_URL }}gui/js/vm/vm-add.js" type="text/javascript"></script>
   <script src="{{ STATIC_URL }}gui/js/obj.js" type="text/javascript"></script>
+  <script src="{{ STATIC_URL }}gui/js/mon/zapi.js" type="text/javascript"></script>
```
dn0: This shouldn't be called `zapi`. Lets name it after the namespace in the API (alerts or alerting).

gui/static/gui/js/mon/zapi.js
```diff
@@ -0,0 +1,54 @@
...
+
+  var self = this;
+
+  var list_table_html = function(msg) {
+    return '<table class="box table table-striped table-condensed table-responsive" id="tasklog-table"><thead></thead>' +
+           '<tbody id="tasklog" class="nowrap"><tr><td colspan="8">' +
+           '<p class="msg">' + msg + '</p></td></tr></tbody><tfoot></tfoot></table>'
+  };
+
+  this.update = function(yyyymm, result) {
```
dn0: What is yyyymm?

gui/static/gui/js/mon/zapi.js
```diff
@@ -0,0 +1,54 @@
...
+
+function mon_ZAPI(view_name, hostname) {
+  this.view_name = view_name;
+  // TODO: add ability to filter by hostname (and other filters, eg. date)
+  this.hostname = hostname;
+
+  var self = this;
+
+  var list_table_html = function(msg) {
+    return '<table class="box table table-striped table-condensed table-responsive" id="tasklog-table"><thead></thead>' +
```
dn0: HTML must go into the template

gui/static/gui/js/mon/zapi.js
```diff
@@ -0,0 +1,54 @@
...
+           '<tbody id="tasklog" class="nowrap"><tr><td colspan="8">' +
+           '<p class="msg">' + msg + '</p></td></tr></tbody><tfoot></tfoot></table>'
+  };
+
+  this.update = function(yyyymm, result) {
+    if (typeof(result) === 'undefined') {
+      // NO result, we dont have task yet!
+      // Create task and wait for callback with result.
+      if (SOCKET.socket.connected) {
```
dn0: You shouldn't need to check the socket.io connection. It's already done in esio().

gui/static/gui/js/gsio.js
```diff
@@ -218,6 +218,19 @@ function mon_get_history(obj_type, hostname, graph, data) {
...
 }
 
+function mon_get_alerts(hostname, data) {
+  var args = [];
+  var kwargs = {'data': {}};
+  //if (typeof(data) !== 'undefined') {
+  //  kwargs['data'] = data;
+  //}
+  // This runs in background, so do not run if not connected
+  if (SOCKET.socket.connected) {
```
dn0: This check is not needed, see esio().

gui/static/gui/js/mon/zapi.js
```diff
@@ -0,0 +1,54 @@
+/*
+ * Universal Zabbix functions (used by mon_alert_list)
+ */
+
+var ZAPI = null;
```
dn0: Do not use the ZAPI name here, or reference Zabbix directly. The JS code connects to our API and our API doesn't necessarily have to connect to Zabbix.

Also create a function/class for alerts/alerting only.

gui/static/gui/js/mon/zapi.js
ricco386: Actually I do, I know that there is check for this in esio, but that is too late (or early depends on point of view ;)).

In my case if user hit refresh in browser on this page he would never got the task results, eg. he would see the dim loading screen (possibly we can change to No alerts default template (without data))...

In else block there is an explanation why he does not see any data after browser refresh, that he should navigate away and return in order to get fresh data.

This was done mostly for the UX if someone refresh...

dn0: Why is this happening after the refresh? - lets fix this.

ricco386: After refresh Socket IO is not connected yet and the javascript is run... It would run mon_get_alerts which would create a task, but only when Socket IO is connected. Since this is not run there is no callback and there are no data (and no code to disable dim screen is run).

The check needs to be there either to show the current message, or can be rewritten to do something else, for example sleep for second or two and try to connect again and so on...

gui/static/gui/js/mon/zapi.js
ricco386: Preparation for filtering... TBD

api/mon/base/serializers.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class AlertSerializer(s.Serializer):
+    """
+    Used by NodeHistoryView to validate zpools value.
+    """
+    since = s.TimeStampField(required=False)
+    until = s.TimeStampField(required=False)
+    last = s.IntegerField(required=False)
+    display_notes = s.BooleanField(default=True)
```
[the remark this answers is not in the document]
ricco386: Updated API doc and simplified params as the naming was not so intuitive. Currently we have ability to have show_events which is default `true` because:
1. List of events related to alert, which is important
2. GUI is listing this data

api/mon/base/serializers.py
ricco386: Serializer pre-set `hosts` parameter to hosts in DC, user can use only subset of hosts in DC otherwise he would have exception. We can also do api call `dc_unbound` that can set call as DC unbound, by default is is false and only super admin can use it and see everything.

api/mon/base/serializers.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def __init__(self, obj=None, instance=None, data=None, **kwargs):
+        self.obj = obj
+        # We cannot set a function as a default argument of TimeStampField
+        if data is None:
+            data = {}
+        else:
+            data = data.copy()
+
+        if 'since' in data and 'until' not in data:
```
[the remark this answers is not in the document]
ricco386: Refactored to validate function. Not entirely sure but if I do not set until it is not set and send to zapi and return weird results, with until it works fine....

api/mon/base/tasks.py
```diff
@@ -11,7 +11,7 @@
 from que.mgmt import MgmtTask
 from vms.models import Dc
 
-__all__ = ('mon_clear_zabbix_cache', 'mon_template_list', 'mon_hostgroup_list')
+__all__ = ('mon_clear_zabbix_cache', 'mon_template_list', 'mon_hostgroup_list', 'mon_alert_list')
```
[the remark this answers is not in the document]
ricco386: refactored

api/mon/base/tasks.py
```diff
@@ -88,3 +88,39 @@ def mon_hostgroup_list(task_id, dc_id, **kwargs):
...
+@cq.task(name='api.mon.base.tasks.mon_alert_list', base=MgmtTask)
+@mgmt_task()
+def mon_alert_list(task_id, dc_id, *args, **kwargs):
+    """
+    Return list of alerts available in Zabbix.
+    """
+    dc = Dc.objects.get_by_id(int(dc_id))
+
+    if is_task_dc_bound(task_id):
+        # TODO: set hosts_or_groups to hosts in this DC.
```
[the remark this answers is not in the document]
ricco386: done in serializer

gui/mon/views.py
ricco386: removed

gui/static/gui/js/mon/alerts.js
```diff
@@ -0,0 +1,59 @@
...
+          }
+        }
+      }
+      hide_loading_screen();
+    }
+  };
+} // mon_Alerts: Zabbix alerts!
+
+function alert_update(filter, result) {
+  MONITORING_ALERTS.update(filter, result);
```
dn0: Add a condition so the update executes only if the the right view is displayed (search for `is_displayed`) in our JS files.

gui/static/gui/js/mon/alerts.js
```diff
@@ -0,0 +1,59 @@
...
+            $("#alert-list-table").html(data);
+          } else {
+            $(".msg").html(textStatus);
+          }
+        }, filter);
+
+      } else {
+        // Something is not alright, loop throught result and alert messages!
+        // This is called on esio FAILED (API error)
+        for (var key in result.result) {
```
dn0: Checkout `_sererror_from_result` in gsio.js if it can be used instead.

gui/static/gui/js/mon/alerts.js
```diff
@@ -0,0 +1,59 @@
...
+
+  this.update = function(filter, result) {
+
+    if (typeof(result) === 'undefined') {
+      // NO result, we dont have task yet!
+      // Create task and wait for callback with result.
+      if (SOCKET.socket.connected) {
+        show_loading_screen('Getting data from Zabbix', true);
+      } else {
+        $(".msg").html(
```
dn0: Selector too generic.

gui/static/gui/js/mon/alerts.js
```diff
@@ -0,0 +1,59 @@
...
+function mon_Alerts(view_url) {
+  this.view_url = view_url;
+
+  this.update = function(filter, result) {
+
+    if (typeof(result) === 'undefined') {
+      // NO result, we dont have task yet!
+      // Create task and wait for callback with result.
+      if (SOCKET.socket.connected) {
+        show_loading_screen('Getting data from Zabbix', true);
```
dn0: Use `gettext()` for user-visible text.

Issue #210, opened 2017-07-25
Automatically sync Zabbix hosts after changing DC or global MON_ settings


dn0:
Lets discuss this first...

After changing `MON_ZABBIX_*_VM` DC settings, the user can select all hosts in a DC a press *Update* to force a zabbix sync.

After changing  `MON_ZABBIX_*_NODE` global settings the situation is different and triggering a zabbix sync is not that straightforward.

1) Maybe we should trigger a zabbix sync after changing any of the zabbix settings.
2) Or we should implement an *Update* button for compute nodes.

PR #309, opened 2017-11-22
Fixed race condition in get_or_create_hostgroups()

dn0:
Related to #210 

- Renamed _get_or_create_groups() -> _get_or_create_hostgroups()
- Added RemoteObjectAlreadyExists exception
- Switched to RemoteObjectDoesNotExist in proper places (backward compatible)
- Added ZabbixUserGroupContainer.get()
- Fixed get_template_list() and get_hostgroup_list()

### DIFF
--- a/api/mon/backends/zabbix/base.py
+++ b/api/mon/backends/zabbix/base.py
@@ -2,10 +2,12 @@ from logging import getLogger, INFO, WARNING, CRITICAL, ERROR
 from time import time
 from operator import itemgetter
 from datetime import datetime
+from functools import wraps
 from subprocess import call
 import re
 
 from django.utils.six import iteritems
+from django.utils.decorators import available_attrs
 from django.db.models import Q
 from frozendict import frozendict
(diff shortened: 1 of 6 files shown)


api/mon/backends/zabbix/base.py
```diff
@@ -933,20 +949,41 @@ def name(self):
     def name(self, value):
         raise ValueError('Name is immutable')
 
+    @staticmethod
+    def __call_zapi(zapi, zapi_method, params=None):
```
secult: Isn't it possible to make this more elegant, perhaps moving only the exception handling to a separate method, or using a decorator?

api/mon/backends/zabbix/base.py
```diff
@@ -1439,11 +1477,18 @@ def hostgroup_name_factory(hostgroup_name, dc_name):
 
         return name
 
+    def get(self):
```
secult: I prefer calling this method "refresh" and return nothing (as in ZabbixUserGroupContainer)

PR #311, opened 2017-11-22
Added automatic synchronization of Zabbix hosts after changing DC settings

### LINKED ISSUE #210: Automatically sync Zabbix hosts after changing DC or global MON_ settings

dn0:
Closes #210


api/dc/base/dc_settings.py
```diff
@@ -45,18 +44,7 @@ def put(self):
         dc.custom_settings = dcs
         dc.save()
         data = ser.data
```
secult: you don't have to reassign this variable to a new name

api/mon/base/tasks.py
```diff
@@ -38,6 +35,37 @@ def mon_clear_zabbix_cache(task_id, dc_id, full=True):
...
+    Clear Zabbix cache and sync everything in Zabbix.
+    Related to a specific DC.
+    Triggered by dc_settings_changed signal.
+    """
+    dc = Dc.objects.get_by_id(int(dc_id))
+
+    if clear_cache:
+        logger.info('Clearing zabbix cache in DC %s', dc)
+        mon_clear_zabbix_cache(dc)
+        get_monitoring(dc)  # Cache new Zabbix instance for tasks below
```
secult: Is this caching useful as the other tasks are run separately?

dn0: @secult Yes it is ... you save calls to the Zabbix API -> look for the `@cache_result@` decorator in api.mon.backends.zabbix. We cache host ids, hostgroup ids, ...

But after changing some monitoring settings we need to clear the cache, e.g. if you change your zabbix server in the settings the IDs are going to change too.

api/dc/base/dc_settings.py
dn0: You're right, I don't have to assign it. But I have to call `ser.data` before calling `ser.detail_dict()`.
I've added a comment.

api/dc/base/dc_settings.py
```diff
@@ -43,7 +43,7 @@ def put(self):
         old_settings = dc.custom_settings
         dc.custom_settings = dcs
         dc.save()
-        data = ser.data
+        data = ser.data  # Prepare ser._data for ser.detail_dict() to work
```
secult: In that case, this is very unintuitive and I would suggest removing this dependency, perhaps by changing api.dc.base.serializers.DcSettingsSerializer#detail_dict to prepare the self._data variable if they doesn't exist. Simple accessing a property should not alter the behavior of seemingly non-related method in such a different way.

secult: This is Django REST Framework rependent and therefore we don't want to break the compatibility

PR #321, opened 2017-12-14
Refactored DictField + Added value size limit to metadata field

dn0:
Related update of user guide: https://github.com/erigones/esdc-docs/commit/9182b993256328503d11d3feb6e6b1125d383fa0


dn0:
@secult In see commit f7c9974 you can see some forgotten code, that is not needed anymore. The idea of the `repr` and `load_repr` was that we wanted to allow copy-paste of jsons into metadata values (which accept only strings). In the mean-time we've introduced raw json editing, which made this code obsolete. If you look at `load_repr()` you'll find a bug -> the string is modified, then parsed and in case of an error the modified string is returned (instead of the original string). 


Obviously this is not the real problem. We are changing VM configuration via `vmadm update` and we pass the json with the update payload via stdin. vmadm calls `fs.readFile('/dev/stdin')` (https://github.com/erigones/smartos-live/blob/master/src/vm/sbin/vmadm.js#L697), which seems OK. But it turns out that `fs.readFile` is broken https://github.com/nodejs/node-v0.x-archive/issues/7412 in nodejs shipped in SmartOS :(

https://github.com/joyent/smartos-live/issues/753

PR #329, opened 2017-12-19
add overlay automation script

YanChii:
What's working:
- creating/updating/deleting overlays on all/subset of nodes
- saving all data into cfgdb to guarantee reproducible changes
- updating config on all nodes without requiring any additional info
- generating IPSec keys, rules, restarting services
- enabling firewall, generating rules, hardening external interfaces

What needs more work
- creating adminoverlay, adding IP addresses to nodes
- altering zabbix agent cfg on nodes?
- IPSec is not tested enough

Dependent on genhosts update in v2.7 branch (for testing, comment out calling `update_ans_hosts` function and create `hosts.cfg` manually).


YanChii:
The shell code isn't polished yet. I'll fix it after I get it work.

YanChii:
Closes erigones/esdc-factory#91

ans/overlays/esdc-overlay.yml
```diff
@@ -0,0 +1,188 @@
...
+  # this will fail loudly if overlay_rules were not parsed correctly
+  - name: fail if config inconsistent
+    fail:
+      msg: "Incorrectly defined input variables! Aborting"
+    run_once: true
+    when:   overlay_rules is undefined or
+            ipsec_keys is undefined
+
+  - name: get node uuid
+    shell: sysinfo -p| grep ^UUID= | cut -d= -f2 | tr -d "'"
```
dn0: Why not use `sysinfo  | json UUID`?

bin/esdc-overlay
```diff
@@ -0,0 +1,665 @@
...
+MGMT_IP=
+DEBUG=0
+# current list of compute nodes from mgmt
+declare -A NODE_DC_LIST		# nodelist: ${!NODE_DC_LIST[@]} ; dc list: ${NODE_DC_LIST[@]}|sort|uniq
+# final overlay config
+ORULES_JSON=
+# final IPSec config
+
+usage() {
+	echo
```
dn0: This would be more editable with `cat << EOF`...

PR #340, opened 2018-01-12
now esdc-overlay scripts generates also mac addresses

YanChii:


YanChii:
Tested, seems working.

Issue #377, opened 2018-07-11
esdc-overlay command fails when DC name contains space


YanChii:
How to replicate:
```
node1# . /usbkey/config
node1# ssh $mgmt_admin_ip /opt/erigones/bin/ctl.sh genhosts --nodes --pdc 'space inside'
Traceback (most recent call last):
  File "/opt/erigones/bin/manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/opt/erigones/envs/lib/python2.7/site-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "/opt/erigones/envs/lib/python2.7/site-packages/django/core/management/__init__.py", line 346, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/opt/erigones/envs/lib/python2.7/site-packages/django/core/management/base.py", line 394, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/opt/erigones/core/management/commands/_base.py", line 97, in execute
    return super(DanubeCloudCommand, self).execute(*args, **options)
  File "/opt/erigones/envs/lib/python2.7/site-packages/django/core/management/base.py", line 445, in execute
    output = self.handle(*args, **options)
TypeError: handle() got multiple values for keyword argument 'vms'
```

YanChii:
The problem seems to be in argument parsing, probably somewhere in `genhosts.py`.
This is a quick workaround without touching the python code:
```
diff --git a/bin/esdc-overlay b/bin/esdc-overlay
index 5703e84..f4711f9 100755
--- a/bin/esdc-overlay
+++ b/bin/esdc-overlay
@@ -144,7 +144,7 @@ update_ans_hosts() {
        fi

        # Update ansible inventory
-       ${SSH} "root@${MGMT_IP}" "${ERIGONES_HOME}/bin/ctl.sh" genhosts --nodes --pdc "${dc_name}" > "${HOSTS_TMP_FILE}"
+       ${SSH} "root@${MGMT_IP}" "${ERIGONES_HOME}/bin/ctl.sh" genhosts --nodes --pdc "'${dc_name}'" > "${HOSTS_TMP_FILE}"
        # shellcheck disable=SC2181
        if [[ $? -eq 0 ]]; then
                # update succeeded
```

PR #378, opened 2018-07-19
Overlay wipe command

YanChii:
This PR adds 2 things:
1. fixes incorrect node matching regex caused probably by differrent sed dialect (not matching tab)
2. adds a wipe command that helps to cleanup/rollback installation of overlay
J.


YanChii:
3. add quotes around physical datacenter name when generating ansible hosts file for esdc-overlay command (#377)

PR #383, opened 2019-01-18
added script for converting VM disks from other virtualization platforms

YanChii:


bin/esdc-convert-disk-to-vm
```diff
@@ -0,0 +1,73 @@
...
+The first attached disk of the destination VM will be overwritten.
+
+Usage:   $0 <image_file> <dest_vm_alias>
+Example: $0 disk1.vmdk mynewvm.lan
+
+EOF
+	exit 1
+fi
+
+if [[ ! -e "${SRC_FILE}" ]]; then
```
dn0: Must by a file - so why not use`-f`?

YanChii: Because this way you can use as a source also zvol from other VM. Useful for e.g. cloning the VM disk using different compression method or dedup.

dn0: 👍

PR #384, opened 2019-01-20
update esmigrate to reflect new presence of vminfod

YanChii:


YanChii:
Some already-pushed commits got here. But the diff is good so I believe squash will solve it.

BTW more reading about `vminfod`: https://github.com/joyent/rfd/blob/master/rfd/0039/README.md

PR #385, opened 2019-01-21
allow upgrading a platform from local custom platform file

YanChii:


bin/eslib/esdc-platform-upgrade-disk.sh
```diff
@@ -174,9 +181,11 @@ fi
 # start download
 printmsg "Downloading the new platform"
 
-# shellcheck disable=SC2086
-if ! ${CURL} ${CURL_QUIET} ${CURL_DEFAULT_OPTS} -o "${PLATFORM_FILE}" "${PLATFORM_DOWNLOAD_URL}"; then
-	die 5 "Cannot download new platform archive. Please check your internet connection."
+if [[ ! -f "${PLATFORM_FILE}" ]]; then
```
dn0: Should we also check some checksum? .e.g. to protect from situation where the file was not fully downloaded...

YanChii: We don't always have a checksum. How about downloading to temporary file first?
See newest commit.

bin/eslib/esdc-platform-upgrade-disk.sh
```diff
@@ -174,9 +181,16 @@ fi
...
+	PLATFORM_TMPFILE="${PLATFORM_FILE}.part"
+	rm -f "${PLATFORM_TMPFILE}"
+	# shellcheck disable=SC2086
+	if ! ${CURL} ${CURL_QUIET} ${CURL_DEFAULT_OPTS} -o "${PLATFORM_FILE}" "${PLATFORM_DOWNLOAD_URL}"; then
```
dn0: Still downloads to original file, no?
Also, now the `if` is inverted - does not need `!` (i.e. harder to read)

YanChii: The point is that now it does not leave unfinished downloads. The final file is created only when the curl returns success.
I'll revert the `if`.

PR #403, opened 2019-02-08
Added manual cleanup script

dn0:
This script can be currently used after a VM migration fails. It will
remove the "_ghost VM_" and also fix the status of the migrated VM
(remove the lock).

Usage:

ssh to mgmt01 and run:
```
/opt/erigones/bin/ctl.sh cleanup vm_migrate hostname=<hostname-of-the-migrated-vm>
```


YanChii:
I thought that ghost VM is created only at replication, not during the migration. I've seen many migration fails without leftovers.

Under which occasion can it be left uncleaned?

YanChii:
Sry for closing, wrong button :)

PR #404, opened 2019-02-12
fix esdc-overlay ADMINOVERLAY_NODE_MAC_MAP

YanChii:
Whithout this, the `esdc-overlay adminoverlay-update` fails to execute.


YanChii:
Ignoring python check fails as this is not applicable to this PR.

PR #406, opened 2019-02-15
fix node version sort during upgrades

### LINKED ISSUE #398: Fix version sort in node upgrade
Node upgrade does not sort correctly versions with suffixes, e.g. `beta1`, `rc2`, etc. This creates side issues with quick mini-releases.

YanChii:
Closes #398


bin/eslib/functions.sh
```diff
@@ -63,6 +63,8 @@ LOFIADM=${LOFIADM:-"/usr/sbin/lofiadm"}
 TAR=${TAR:-"/usr/bin/tar"}
 DD=${DD:-"/usr/bin/dd"}
 FSTYP=${FSTYP:-"/usr/sbin/fstyp"}
+[[ "$( uname -s )" != "Linux" ]] && GSORT=${GSORT:="/opt/local/bin/gsort"}
```
dn0: An `if/else` statement would be more readable here.

Issue #408, opened 2019-02-17
Upgrade to 4.0 beta - ansible issue


YanChii:
During upgrade, the command 
```/opt/erigones/bin/ctl.sh deploy --update```
fails to upgrade ansible because it needs to have `PyNaCl` installed first.

New installations have solved it by placing this requirement before ansible install:
https://github.com/erigones/esdc-ce/pull/407/files#diff-97d6fce8f1d9759e88d0fea21685e653R21

Solution for upgrades can be one of these:
1. uninstall ansible prior to upgrading (if version <= `v3.0.0`, then `pip uninstall ansible`)
2. install `PyNaCl` before upgrading requirements

I'm not sure however where to put this code.

Any ideas?

dn0:
This problem is not related to `PyNaCl`, going to debug this...

dn0:
@YanChii Here is a potential fix - #417 (needs testing)

PR #416, opened 2019-02-24
fix ansible upgrade

YanChii:


YanChii:
> Does the problem with uninstallation repeat itself after a new version is installed?
Yes

YanChii:
Currently, this is the only blocker for `v4.0-beta1` release. 
Personally, I'd rather see this code in `ctl.sh deploy --update` part and not here. But it doesn't change the workaround itself.
And currently I don't see any better way for this (besides of drop-in-replacing of whole `envs/`).

dn0:
@YanChii I understand that this is a blocker and in general I like simple and pragmatic solutions.
However, here I'm against this solution because we don't understand the initial problem well enough. It could potentially lead to other problems (e.g. affecting upgrades of other packages).
I'm going back to issue #408 and continue the debugging there.

PR #417, opened 2019-02-25
Fixed pip uninstall by setting TMPDIR to the same FS as erigones/envs

### LINKED ISSUE #408: Upgrade to 4.0 beta - ansible issue

dn0:
Closes #408

### DIFF
--- a/bin/ctl.sh
+++ b/bin/ctl.sh
@@ -54,6 +54,20 @@ set_django_settings() {
 	export DJANGO_SETTINGS_MODULE
 }
 
+update_env() {
+	export PYTHONUNBUFFERED="true"
+
+	case "${1}" in
+		pip_*)
+			# When uninstalling or upgrading Python packages with symlinks our version of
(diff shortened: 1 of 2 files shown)


dn0:
Requires testing.

YanChii:
Tested by upgrading from `v3.0.0` and working.

PR #425, opened 2019-03-07
modify upgrade to v3.0.0 to match newly installed systems

YanChii:
Maybe not important anymore. But for sake of correctness...


YanChii:
No we don't. And that is the problem. Newly installed systems [do not have](https://github.com/erigones/esdc-factory/blob/master/ansible/vars/build/os/esdc-mgmt.yml#L72) this `eth0` condition in the entry (and that is correct).

PR #426, opened 2019-03-07
Fix ha scripts

YanChii:
This was a lot of work but we have a functional HA again. 

The major change is to switch from unsupported `pcs` library to a new one. It required the changes in the library itself.

The rest is mainly adapting to new ansible version and fixing some bugs.


YanChii:
@dn0 how can we switch off travis warnings about the files in the role's library? I don't indent to fix foreign code.

PR #431, opened 2019-04-15
fix esrep failover/resync

### LINKED ISSUE #409: esrep VM failover error on new platform
On a new platform (201902), the replication VM failover gives the following error:
```
svcadm: Restarter for instance "svc:/application/esrep-sync:slave-replica1" is unavailable.
```
But on the destination node itself, the service name is
```
svc:/application/esrep-sync:slave-436bc567-bc29-406c-9d4f-dea419d7bf83
```
We need to find the source of this inconsistency.

This problem is not present on older platform (2018).

YanChii:
Fixes #409

### DIFF
--- a/bin/eslib/esrep.py
+++ b/bin/eslib/esrep.py
@@ -772,6 +772,7 @@ timeout_seconds='60'>
             self._vm_disks_cleared.append((dst_disk, dst_disk_property))
             self._clear_dataset_property(dst_disk, 'readonly')
 
+        self._vm_start(self.dst_uuid)
         self._vm_disks_cleared = []  # Start was successful
 
         return {
--- a/bin/eslib/functions.sh
+++ b/bin/eslib/functions.sh
@@ -690,19 +690,19 @@ _vm_start() {
 _vm_stop() {
 	local uuid="$1"
 
-	# stopping the stopped VM returns failure
-	if [[ "$(_vm_status "${uuid}")" != "stopped" ]]; then
-		${VMADM} stop "${uuid}" 2>&1
-	fi
+	# stopping the stopped VM returns failure anyway
+	[[ "$(_vm_status "${uuid}")" == "stopped" ]] && return 0
+
+	${VMADM} stop "${uuid}" 2>&1
 }
 
 _vm_stop_force() {
 	local uuid="$1"
 
-	# stopping the stopped VM returns failure
-	if [[ "$(_vm_status "${uuid}")" != "stopped" ]]; then
-		${VMADM} stop "${uuid}" -F 2>&1
-	fi
+	# stopping the stopped VM returns failure anyway
+	[[ "$(_vm_status "${uuid}")" == "stopped" ]] && return 0
+
+	${VMADM} stop "${uuid}" -F 2>&1
 }
 
 _vm_destroy() {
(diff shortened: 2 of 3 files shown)


YanChii:
Please check especially [this](https://github.com/erigones/esdc-ce/pull/431/files#diff-694c4243cf0065c7fe6800f0a942483fR775) line as I'm not sure if this is a fix on the right place. 
But the thing is that dst VM really did not get started after failover.


Also we should put somewhere to the docs an example of failover using API:
```
es put /vm/<vm_name>/replica/<replica_name>/failover
e.g:
es put /vm/centos-test.lan/replica/replica11/failover
```
Because when the src node becomes unreachable, the control buttons for a VM get hidden and consequently the failover cannot be executed from the GUI. We might consider calling this one also a bug.

bin/eslib/functions.sh
```diff
@@ -690,13 +690,19 @@ _vm_start() {
 _vm_stop() {
 	local uuid="$1"
 
-	${VMADM} stop "${uuid}" 2>&1
+	# stopping the stopped VM returns failure
+	if [[ "$(_vm_status "${uuid}")" != "stopped" ]]; then
```
dn0: Not strictly necessary, but I think this check could be more readable if it would be written as a "guard" before the vmadm stop command, i.e.:
```
[[ "$(_vm_status "${uuid}")" != "stopped" ]] && return 0
```

bin/eslib/functions.sh
```diff
@@ -690,13 +690,19 @@ _vm_start() {
...
+		${VMADM} stop "${uuid}" 2>&1
+	fi
 }
 
 _vm_stop_force() {
 	local uuid="$1"
 
-	${VMADM} stop "${uuid}" -F 2>&1
+	# stopping the stopped VM returns failure
+	if [[ "$(_vm_status "${uuid}")" != "stopped" ]]; then
```
dn0: Similar as above, "guard" style early exit from function would look better here.

bin/eslib/functions.sh
```diff
@@ -967,12 +973,20 @@ _service_disable() {
 	local fmri="$1"
 
 	${SVCADM} disable -s "${fmri}" && _service_file_save "${fmri}"
+
+	# If there's any SMF service in maintenance state,
+	# svcadm disable with "-s" returns a non-zero code.
+	# Therefore we have to explicitly check and return
+	# the result of this operation.
+	[[ "$(_service_status "${fmri}")" == "disabled" ]]
```
dn0: I like this, however in this case the `_service_file_save "${fmri}"` is not triggered -> the updated service manifest is not saved to disk. Please fix this this.

bin/eslib/functions.sh
```diff
@@ -967,12 +973,20 @@ _service_disable() {
...
 }
 
 _service_enable() {
 	local fmri="$1"
 
 	${SVCADM} enable -s "${fmri}" && _service_file_save "${fmri}"
+	# see comment in _service_disable
+	[[ "$(_service_status "${fmri}")" == "disabled" ]]
```
dn0: The same problem as above.

bin/eslib/esrep.py
```diff
@@ -772,6 +772,7 @@ def failover(self):
             self._vm_disks_cleared.append((dst_disk, dst_disk_property))
             self._clear_dataset_property(dst_disk, 'readonly')
 
+        self._vm_start(self.dst_uuid)
```
dn0: Makes me wonder why this was not here - maybe we expect the esdc API (callback task) to do this for us?

YanChii: I've looked into esdc API code (started by `VmReplicaFailover()` in `vm_replica_action.py`) and I did not find any VM start.

Logically, it makes more sense to me when the start command resides in the esrep command itself because it should not be dependent on DC mgmt.

I still don't understand why it was not there but the real-world tests show that without this line, the VM doesn't start. So something was missing anyway.

dn0: It is in the API, but in the callback: https://github.com/erigones/esdc-ce/blob/master/api/vm/replica/tasks.py#L355 - Once everything is OK, the callback would have started it. But lets keep it also here. I think it should be ok like this.

YanChii:
Rebased against master to allow merge.

PR #435, opened 2019-05-04
Added VM status refresh button in VM details page

### LINKED ISSUE #434: Allow starting the already-started VM
When starting the VM that is for some reason running on a node but DC thinks it is stopped, it throws following error:
```
Got bad return code (0). Error: Unable to start VM: VM is already 'running'
```
It would be nicer to not throw error, just adjust the state in DC database. Because the desired state is already present. In current state, there's not many options how to recover from inconsistency between DB and current VM state (without using API or ssh).

The same probably applies on the opposite - stop vs stopped state (when DC thinks the VM is running).

dn0:
Closes #434
Updated PUT vm_status action:current to be work only for specific VM states.


dn0:
@YanChii This change is visible on my.erigones.com

YanChii:
I've tried it also on the HZ test and it works.
Perfect!

Issue #439, opened 2019-05-17
Cannot update VM memory


YanChii:
In `v4.0` when modyfying the RAM parameter for a VM, the VM update fails with
```
Got bad return code (1). Error: first of 1 error: vminfod watchForChanges "VM.js update (mytestvm.lan)" timeout exceeded  
```
It seems it's caused by different computation of certain VM json parameters.
This is example of DC-pushed params:
```
{
  "ram": 2048,
  "max_physical_memory": 2304,
  "max_swap": 4096
}
```
And this is what vminfo expects:
```
{
  "ram": 2048,
  "max_physical_memory": 3072,
  "max_swap": 4864
}
```
When I manually push this second json, the VM update succeeds.
It also succeeds with only
```
{
  "ram": 2048
}
```
because it sets the other parameters accordingly.

It will probably be enough to modify DC parameters computation to match the `vminfo` style.

YanChii:
@dn0 I'd need a hand with this one.

PR #440, opened 2019-05-20
fix updating KVM VM memory

YanChii:
I've found the source of inconsistency. The reason is inconsistent setting of `max_locked_memory`.

Reference for similar future debugs:
https://github.com/joyent/triton/blob/master/docs/operator-guide/vminfod/troubleshooting.md


YanChii:
Closes #439

PR #442, opened 2019-05-21
Change default es api_url

YanChii:


YanChii:
This es is the main binary present in the PATH on mgmt01 and nodes. So it's quite confusing. I always struggle to get the api_url right.

I understand the devel purposes. But this is production.

Maybe we can find other way. Do you have some idea?

I'm also fine with not changing it. But maybe we should then move it out of the PATH.

dn0:
@YanChii I don't understand. The default URL also works on mgmt01 in production.

YanChii:
My fault, I didn't expect this url is working in production :) .
Closing.

PR #451, opened 2019-09-22
Fixed wrong action condition, which prevented file backup operations.

klebed:
Fixed an issue with file backup operations.
image

### DIFF
--- a/bin/esbackup
+++ b/bin/esbackup
@@ -112,7 +112,7 @@ def main():
                             fsfreeze=args.fsfreeze)
     elif cmd == 'ds-delete':
         res = bkp.ds_delete(args.snapshot, last_snapshots=args.remote_snapshot, metadata=args.metadata)
-    elif cmd == 'ds-restore' or 'snap-restore':
+    elif cmd == 'ds-restore' or cmd == 'snap-restore':
         res = bkp.ds_restore(args.snapshot, args.destination)
     elif cmd == 'file-create':
         res = bkp.file_create(args.source, args.file, metadata=args.metadata, json=args.json,


YanChii:
Thank you!

YanChii:
FYI if you want to update to this fix immediately, put this commit number (ccb2b023e7d9d1c868f5195aee7ad3b0b646ff4d) instead of the version number when updating (yes, `ccb2b023e7d9d1c868f5195aee7ad3b0b646ff4d` instead of `v4.1`).

In this case, it is enough to update the node software, no need for mgmt upgrade.

Jan

klebed:
> FYI if you want to update to this fix immediately, put this commit number ([ccb2b02](https://github.com/erigones/esdc-ce/commit/ccb2b023e7d9d1c868f5195aee7ad3b0b646ff4d)) instead of the version number when updating (yes, `ccb2b023e7d9d1c868f5195aee7ad3b0b646ff4d` instead of `v4.1`).
> 
> In this case, it is enough to update the node software, no need for mgmt upgrade.
> 
> Jan

Cool, thanks!
But we use USB-boot, should it be built into image and applied to flash-drives?

YanChii:
Danube Cloud code is independent from platform code and you can upgrade it without touching the platform.
- DC code can be ugraded from GUI/API (or by `esdc-git-upgrade`).
- Platform is upgraded only from cmd line by `esdc-platform-upgrade`.

klebed:
> Danube Cloud code is independent from platform code and you can upgrade it without touching the platform.
> 
>     * DC code can be ugraded from GUI/API (or by `esdc-git-upgrade`).
> 
>     * Platform is upgraded only from cmd line by `esdc-platform-upgrade`.

Yeah! Missed that thing! Platform and DC on nodes has separate version data.
Thanks!

YanChii:
And you can test also the other fix you have reported about that dc_bound DNS. But for that, you need to commit-upgrade the mgmt.

PR #499, opened 2020-08-29
fix zabbix template for checking newer pgsql

YanChii:


YanChii:
Not necessary to merge into `v4.4` as the PG upgrade script is still WIP. 
New `t_zabbix-db.json` is present on new installations in Zabbix from DC `v4.3`. The `t_svc-db-ha.json` is unchanged and maybe doesn't need update at all.

